PR #9, opened 2024-10-18
Discover MNNVL topology with single blocksize

ritikasrivastava:
topology.conf output
BlockName = <clusterUUID> Nodes = <comma separated list of node names on the same NVL Domain>

### DIFF
--- a/pkg/providers/baremetal/mnnvl.go
+++ b/pkg/providers/baremetal/mnnvl.go
@@ -62,7 +62,7 @@ func getClusterOutput(ctx context.Context, domainMap map[string]domain, nodes []
 	}
 	return nil
 }
-func toSlurm(domainMap map[string]domain) *common.Vertex {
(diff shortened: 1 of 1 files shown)


pkg/providers/baremetal/mnnvl.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		nodeMap := domainMap[clusterUUID].nodeMap
+		nodeMap[nodeName] = true
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("Scanner error while reading pdsh output\n")
+	}
+	return nil
+}
+func toSlurm(domainMap map[string]domain) *common.Vertex {
```
dmitsh: nit: rename "toSlurm" to "toGraph" or something similar.

ritikasrivastava: > nit: rename "toSlurm" to "toGraph" or something similar.

Addressed

PR #15, opened 2024-10-24
Topograph model import for use with toposim

henryh2:
This MR adds functionality for topograph to import a cluster model from a file to be use with toposim.


cmd/topograph/main.go
```diff
@@ -27,15 +27,18 @@ import (
...
 )
 
 var GitTag string
 
 func main() {
 	var c string
+	var m string
 	var version bool
 	flag.StringVar(&c, "c", "/etc/topograph/topograph-config.yaml", "config file")
+	flag.StringVar(&m, "m", "", "model file to use when using the test provider")
```
dmitsh: Suggestion: instead of passing model as a command line argument to topograph, pass it as a parameter in topology request. That way we could run different models without restarting topograph

dmitsh (changes requested):

dmitsh (approved):

pkg/factory/provider.go
```diff
@@ -65,11 +66,25 @@ type testProvider struct {
...
+	if len(params) != 0 {
+		modelPath = params[common.KeyModelPath]
+	}
+
+	if len(modelPath) == 0 {
+		p.tree, p.instance2node = translate.GetTreeTestSet(false)
+	} else {
+		model, err := models.NewModelFromFile(modelPath)
+		if err != nil {
+			return nil, err
```
pkedy, line 82: nit: You unless properly wrapped by `models.NewModelFromFile`, you'll want to wrap this error with

```go
fmt.Errorf("could not read model file: %w", err)
```

If `models.NewModelFromFile(modelPath)` DOES wrap the error, I would just add a comment

```go
return nil, err // Wrapped by models.NewModelFromFile
```

pkedy (commented):
We need to start putting more comments on public funcs. One small nit.

PR #18, opened 2024-10-29
Blocksize refactoring

ritikasrivastava:
1. Domain sizes may vary
2. Return block size as largest power of 2, less than the minimum domain size
3. Verify admin provided blocksize > 0 and <= minimum domain size


pkg/translate/output.go
```diff
@@ -79,16 +80,24 @@ func printDisconnectedBlocks(wr io.Writer, root *common.Vertex, domainVisited ma
...
+	minDomainSize := -1
+	for _, dSize := range domainVisited {
+		if minDomainSize == -1 || minDomainSize > dSize {
+			minDomainSize = dSize
 		}
 	}
-	return blockSize
+	if adminBlockSize != "" {
+		blockSizes := strings.Split(adminBlockSize, ",")
+		planningBS, _ := strconv.Atoi(blockSizes[0])
```
dmitsh: I would add checking error here, in case the input is not valid, in which case we can ignore provided block size

pkg/translate/output.go
```diff
@@ -79,16 +80,24 @@ func printDisconnectedBlocks(wr io.Writer, root *common.Vertex, domainVisited ma
...
+func getBlockSize(domainVisited map[string]int, adminBlockSize string) string {
+	minDomainSize := -1
+	for _, dSize := range domainVisited {
+		if minDomainSize == -1 || minDomainSize > dSize {
+			minDomainSize = dSize
 		}
 	}
-	return blockSize
+	if adminBlockSize != "" {
+		blockSizes := strings.Split(adminBlockSize, ",")
```
dmitsh, line 91: check that len(blockSezes) == 2 ?

ritikasrivastava: If ',' is not present, then Split will return entire adminBlockSize which means only 1 blockSize is entered by the admin. In either case, we need access to only blockSizes[0]. So checking for length 2 is not necessary IMO

dmitsh: ok, fair

pkg/translate/output.go
ritikasrivastava: Added.

pkg/translate/output.go
```diff
@@ -79,16 +80,28 @@ func printDisconnectedBlocks(wr io.Writer, root *common.Vertex, domainVisited ma
...
+		planningBS, err := strconv.Atoi(blockSizes[0])
+		if err != nil {
+			fmt.Printf("Alert, strconv Atoi for admin provided blockSize %v failed with err: %v! Ignoring it\n", blockSizes[0], err)
```
dmitsh, line 94: since we do not return error, could you add Prometheus metrics to monitor this issue?

pkg/translate/output.go
```diff
@@ -79,16 +80,28 @@ func printDisconnectedBlocks(wr io.Writer, root *common.Vertex, domainVisited ma
...
+		} else {
+			if planningBS > 0 && planningBS <= minDomainSize {
+				return adminBlockSize
+			}
+			fmt.Printf("Alert Overriden planning blockSize of %v does not meet criteria, minimum domain size %v! Ignoring it\n", planningBS, minDomainSize)
```
dmitsh, line 99: same metrics can be used here.

dmitsh: I will implement the metrics in a separate PR

PR #21, opened 2024-10-30
add Prometheus metrics for nodes with missing topology

dmitsh:


pkg/providers/aws/instance_topology.go
```diff
@@ -184,7 +186,7 @@ func toGraph(topology []types.InstanceTopology, cis []common.ComputeInstances) (
 				ID:   instanceID,
 			}
 		}
-		forest["cpu-nodes"] = sw
+		forest["no-topology"] = sw
```
henryh2: So is this "no-topology" vertex going to be a hard-coded switch for tree topology? In that case, should we placed it in the consts file in common and edit some of the tests (such as in grpc_client_test.go) to follow the same strucutre?

pkg/providers/oci/instance_topology.go
```diff
@@ -217,7 +219,7 @@ func toGraph(bareMetalHostSummaries []*core.ComputeBareMetalHostSummary, cis []c
 				ID:   instanceID,
 			}
 		}
-		forest["cpu-nodes"] = sw
+		forest["no-topology"] = sw
```
henryh2: Same as previous comment with AWS nodes.

dmitsh: agree

dmitsh: done

XRFXLP (approved):
LGTM

PR #26, opened 2024-11-02
Require metadata on all topology

henryh2:
Requires metadata to be set on all topology, including both tree and block topology.


henryh2:
Two comments regarding this PR, the first of which is:

If the topology request payload specifies a topology plugin (either tree or block) that is different than the topology that can be provided, what should be the action? Return an error in the request?

henryh2:
And second, for MNNVL, is block topology always returned (current implementation), or should tree topology be returned in some situations?

pkg/models/model.go
```diff
@@ -26,6 +26,7 @@ import (
 )
 
 type Model struct {
+	Topology       string          `yaml:"topology"`
```
dmitsh: Why do we need Topology field here?

dmitsh:
> If the topology request payload specifies a topology plugin (either tree or block) that is different than the topology that can be provided, what should be the action? Return an error in the request?

Toposim (and GTS) returns a collections of nodes with their local topology (a chain of switches and/or presence of nvlink).
From here we can construct either tree or block topology.  So, IIUC, it is up to the requester to define. IMO we should consider tree topology as default.

pkg/models/model.go
henryh2: The reason is in the GetTestProvider() function in pkg/factory/provider.go. In the case of using a specific model file, which gets imported to file, that tree has to have a metadata tag so that the ToGraph() function can be called on line 82 (if it doesn't, one of the changes is that an error is thrown). So the "Topology" field is necessary in the model file so that the model can be correctly tagged in this context.

pkg/ib/ib.go
```diff
@@ -65,7 +65,14 @@ func GenerateTopologyConfig(data []byte) (*common.Vertex, error) {
 	}
 	seen = make(map[int]map[string]*Switch)
 	root.simplify(root.getHeight())
-	return root.toGraph()
+	rootNode, err := root.toGraph()
+	if err != nil {
```
dmitsh, line 68: other way around?
if err != nil return err

pkg/models/model.go
henryh2: Topology field removed, the model import will now set to block topology if an nvlink domain is specified.

pkg/ib/ib.go
henryh2: Fixed

pkg/models/model.go
```diff
@@ -196,11 +200,13 @@ func (model *Model) ToTree() (*topology.Vertex, map[string]string) {
...
 	}
-	root := &topology.Vertex{
-		Vertices: map[string]*topology.Vertex{
-			topology.ValTopologyBlock: blockRoot,
-			topology.ValTopologyTree:  treeRoot,
-		},
+	if block_topology {
+		root := &topology.Vertex{
+			Vertices: map[string]*topology.Vertex{topology.TopologyBlock: blockRoot, topology.TopologyTree: treeRoot},
+			Metadata: map[string]string{topology.KeyPlugin: topology.TopologyTree},
```
ritikasrivastava: TopologyBlock instead

henryh2: Yes, thank you

ritikasrivastava (approved):

PR #31, opened 2024-11-13
Traverse Topology in DFS format in block topology output

ritikasrivastava:
Implemented on top of metadata-fix PR

DO NOT MERGE before metadata-fix branch. Just for review


henryh2:
So curiously, what is the purpose of having both a BFS and DFS traversal for block topology? Do we need both, or is mostly for testing purposes?

ritikasrivastava:
> So curiously, what is the purpose of having both a BFS and DFS traversal for block topology? Do we need both, or is mostly for testing purposes?

Yes, mainly for testing and review. One of the topology traversals maybe removed as per the review.

PR #32, opened 2024-11-15
Traverse Topology in DFS format in block topology output

ritikasrivastava:


henryh2:
Perhaps this would be a good MR to introduce a document in repo that explains the vertex structure, just to make sure everyone is on the same page. I think that might be pretty helpful, perhaps we should discuss during review?

pkg/translate/output.go
```diff
@@ -111,41 +111,69 @@ func getBlockSize(domainVisited map[string]int, adminBlockSize string) string {
 	return strconv.Itoa(int(bs))
 }
 
-func toBlockTopology(wr io.Writer, root *topology.Vertex) error {
-	// traverse tree topology and when a node is reached, check within blockRoot for domain and print that domain.
+/*
```
dmitsh: Do we need to keep the commented-out code?

pkg/translate/output.go
```diff
@@ -111,41 +111,69 @@ func getBlockSize(domainVisited map[string]int, adminBlockSize string) string {
...
+		blockSize := ""
+		if _, exists := root.Metadata[topology.KeyBlockSizes]; exists {
+			blockSize = root.Metadata[topology.KeyBlockSizes]
+		}
+		blockSize = getBlockSize(domainVisited, blockSize)
+		_, err = wr.Write([]byte(fmt.Sprintf("BlockSizes=%s\n", blockSize)))
+		return err
+	}
+*/
+func toBlockTopologyDFS(wr io.Writer, root *topology.Vertex) error {
```
dmitsh: are we planning to have other types of `toBlockTopology` beside DFS? If not, perhaps we could keep `toBlockTopology()`?

pkg/translate/output.go
ritikasrivastava: No, we do not. Just for review purposes. I can remove it once everyone got a chance to review.

pkg/translate/output.go
ritikasrivastava: No, we do not. Just for review purposes. I can rename it

dmitsh (approved):

dmitsh (approved):

PR #36, opened 2024-11-20
simplify topology model

dmitsh:


pkg/models/model.go
```diff
@@ -141,80 +129,28 @@ func (m *Model) setNodeMap() error {
 	return nil
 }
 
-func getNetworkLayers(name string, swmap map[string]string) ([]string, error) {
-	sw := make(map[string]bool)
-	res := []string{}
+func getNetworkLayers(sw *Switch, swmap map[string]*Switch) ([]string, map[string]string, error) {
+	visited := make(map[string]bool)
```
pkedy: Nit: For `visited`, you need a Set[string] so this could be `make(map[string]struct{})`

pkg/providers/providers_sim.go
```diff
@@ -0,0 +1,39 @@
...
+
+package providers
+
+import (
+	"fmt"
+
+	"github.com/NVIDIA/topograph/internal/config"
+)
+
+type SimParams struct {
```
pkedy: Nit: Name as `SimulationParams`?

pkg/models/model.go
```diff
@@ -141,80 +129,28 @@ func (m *Model) setNodeMap() error {
...
+	layers := []string{}
+	metadata := make(map[string]string)
+
 	for {
+		name := sw.Name
 		// check for circular switch topology
-		if _, ok := sw[name]; ok {
-			return nil, fmt.Errorf("circular topology for switch %q", name)
+		if _, ok := visited[name]; ok {
+			return nil, nil, fmt.Errorf("circular topology for switch %q", name)
```
pkedy: Nit: A little more easy to understand error message.
"circular dependency detected in topology for switch"

pkedy (dismissed):

PR #37, opened 2024-11-22
add Prometheus metrics for block size calculation errors

dmitsh:


pkg/metrics/metrics.go
```diff
@@ -51,12 +51,22 @@ var (
 		},
 		[]string{"provider"},
 	)
+
+	blockSizeErrorsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name:      "blocksize_error_total",
+			Help:      "Total number of blocksize calculation errors.",
```
ritikasrivastava: Calculation errors? or bad admin inputs?

ritikasrivastava (approved):

PR #38, opened 2024-11-23
add unit test for mocking CSP API

dmitsh:


pkg/server/http_server_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func getAvailablePort() (int, error) {
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, err
+	}
+	defer listener.Close()
+
+	return listener.Addr().(*net.TCPAddr).Port, nil
+}
```
pkedy, line 42: Why didn't you use the `httptest` package?

pkedy (approved):
LGTM; however I think you should have used the `httptest` package. :)

pkg/server/http_server_test.go
dmitsh: I tried to use it, but for this test I needed to test server.HttpServer object, which has http.Server as a member.
The httptest.Server exposes different interface, so it is not compatible with http.Server

PR #41, opened 2024-11-28
deprecate 'test' engine

dmitsh:
1. Deprecate 'test' engine and use 'slurm' instead.
2. Replace 'skip_reload' with 'reconfigure' in the engine parameters and flip the logic. By default we won't call `scontrol reconfigure`, which lets us reuse 'slurm' engine for the tests. Only when 'reconfigure' is set to 'true', we will restart slurm scheduler.


pkedy (dismissed):

pkg/engines/slurm/slurm.go
```diff
@@ -140,55 +141,52 @@ func GenerateOutput(ctx context.Context, tree *topology.Vertex, params map[strin
 
 func GenerateOutputParams(ctx context.Context, tree *topology.Vertex, params *Params) ([]byte, error) {
 	buf := &bytes.Buffer{}
-	path := params.TopoConfigPath
+	path, plugin := params.TopoConfigPath, params.Plugin
+
+	// set and validate plugin
+	switch plugin {
+	case "":
+		plugin = topology.TopologyTree
```
ritikasrivastava, line 149: If no plugin is passed, shouldn't we check if block subtree exists? If it does, default should be blockTopology.

pkg/engines/slurm/slurm.go
```diff
@@ -140,55 +141,52 @@ func GenerateOutput(ctx context.Context, tree *topology.Vertex, params map[strin
...
+	case topology.TopologyTree, topology.TopologyBlock:
+		// nop
```
ritikasrivastava: Also, if params provided plugin is Block but block subtree doesn't exist, unsupported message should be reported.

pkg/metrics/metrics.go
```diff
@@ -52,9 +52,9 @@ var (
 		[]string{"provider"},
 	)
 
-	blockSizeValidationErrorsTotal = prometheus.NewCounterVec(
+	validationErrorsTotal = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name:      "blocksize_error_total",
+			Name:      "validation_error_total",
 			Help:      "Total number of blocksize validation errors.",
```
ritikasrivastava: Maybe remove block size word here as well?

pkg/engines/slurm/slurm.go
```diff
@@ -140,55 +141,52 @@ func GenerateOutput(ctx context.Context, tree *topology.Vertex, params map[strin
...
+	default:
+		klog.Infof("Unsupported topology plugin %s. Using %s", plugin, topology.TopologyTree)
+		plugin = topology.TopologyTree
+		metrics.AddValidationError("unsupported plugin")
```
ritikasrivastava, line 164: Maybe create a const for error types?

pkg/translate/output.go
```diff
@@ -30,17 +30,17 @@ import (
...
-		if root.Metadata[topology.KeyPlugin] == topology.TopologyBlock {
-			return toBlockTopology(wr, root)
-		} else if root.Metadata[topology.KeyPlugin] == topology.TopologyTree {
-			return toTreeTopology(wr, root)
-		}
+		plugin = root.Metadata[topology.KeyPlugin]
 	}
-	if root.Vertices[topology.TopologyBlock] != nil {
+
+	if plugin == topology.TopologyBlock {
```
ritikasrivastava, line 40: If Metadata is nil, plugin would be undefined.

pkg/translate/output.go
```diff
@@ -103,13 +103,13 @@ func getBlockSize(domainVisited map[string]int, adminBlockSize string) string {
...
 		if err != nil {
-			metrics.AddBlockSizeValidationError("parsing error")
+			metrics.AddValidationError("block size parsing error")
 			klog.Warningf("Failed to parse blockSize %v: %v. Ignoring.", blockSizes[0], err)
 		} else {
 			if planningBS > 0 && planningBS <= minDomainSize {
 				return adminBlockSize
 			}
-			metrics.AddBlockSizeValidationError("bad domain size")
+			metrics.AddValidationError("bad block domain size")
```
ritikasrivastava, line 112: Same comment,  maybe define certain error codes as const

pkg/translate/output.go
```diff
@@ -476,6 +476,7 @@ func GetBlockWithMultiIBTestSet() (*topology.Vertex, map[string]string) {
 	root := &topology.Vertex{
 		Vertices: map[string]*topology.Vertex{topology.TopologyBlock: blockRoot, topology.TopologyTree: treeRoot},
 		Metadata: map[string]string{
+			topology.KeyPlugin:     topology.TopologyBlock,
```
ritikasrivastava, line 479: If block tree exists, block topology should be the default?

pkg/engines/slurm/slurm.go
dmitsh: For Kubernetes, block topology won't provide enough information.
For Kubernetes we always want to provide tree topology, because we are labeling the nodes according to switch hierarchy.

pkg/engines/slurm/slurm.go
dmitsh: addressed

pkg/metrics/metrics.go
dmitsh: addressed

pkg/engines/slurm/slurm.go
ritikasrivastava: nit: will address in another PR

pkg/translate/output.go
ritikasrivastava: addressed

pkg/translate/output.go
ritikasrivastava: nit: will address in another PR

pkg/translate/output.go
ritikasrivastava: addressed, clarified. tree topology should be the default

PR #47, opened 2024-12-16
Fix tree topology output

ritikasrivastava:


pkg/providers/baremetal/mnnvl.go
```diff
@@ -95,67 +93,96 @@ func getIbTree(ctx context.Context, _ []string) (*topology.Vertex, error) {
 						ibKey := ibPrefix + strconv.Itoa(ibCount)
 						treeRoot.Vertices[ibKey] = ibRoot
 						break
+					} else {
+						fmt.Printf("Missing ibnetdiscover output\n")
```
dmitsh, line 100: Should we use log instead?
Do we need to return error?
Do we need to have a metric for this error?

pkg/providers/baremetal/mnnvl.go
```diff
@@ -95,67 +93,96 @@ func getIbTree(ctx context.Context, _ []string) (*topology.Vertex, error) {
...
 			prefix = tuple[0]
+			// example : nodename-1-[001-004
 			if strings.Contains(tuple[1], "-") {
 				nr := strings.Split(tuple[1], "-")
-				start, _ := strconv.Atoi(nr[0])
-				end, _ := strconv.Atoi(nr[1])
+				w := len(nr[0])
+				start, err := strconv.Atoi(nr[0])
+				if err != nil {
+					fmt.Printf("Atoi err for range start\n")
```
dmitsh: same here

pkg/providers/baremetal/mnnvl.go
```diff
@@ -95,67 +93,96 @@ func getIbTree(ctx context.Context, _ []string) (*topology.Vertex, error) {
...
+				}
+				end, err := strconv.Atoi(nr[1])
+				if err != nil {
+					fmt.Printf("Atoi err for range end\n")
```
dmitsh: same here

pkg/providers/baremetal/mnnvl.go
```diff
@@ -95,67 +93,96 @@ func getIbTree(ctx context.Context, _ []string) (*topology.Vertex, error) {
...
 				} else if strings.Contains(entry, "-") { // suffix containing range of nodes
+					// example: 100-102]
 					nr := strings.Split(entry, "-")
-					start, _ := strconv.Atoi(nr[0])
-					end, _ := strconv.Atoi(nr[1])
+					w := len(nr[0])
+					start, err := strconv.Atoi(nr[0])
+					if err != nil {
+						fmt.Printf("Atoi err for range start when prefix is set\n")
+					}
```
dmitsh, line 166: same here

pkg/providers/baremetal/mnnvl.go
```diff
@@ -95,67 +93,96 @@ func getIbTree(ctx context.Context, _ []string) (*topology.Vertex, error) {
...
+					end, err := strconv.Atoi(nr[1])
+					if err != nil {
+						fmt.Printf("Atoi err for range end when prefix is set\n")
```
dmitsh: same here

pkg/providers/baremetal/mnnvl.go
ritikasrivastava: I am not sure about this. It is not really an error scenario if IB is not present. This text is supposed to be informational. Should we remove it instead?

pkg/providers/baremetal/mnnvl.go
ritikasrivastava: Done

dmitsh (approved):

PR #48, opened 2024-12-19
Add support for cliques

ritikasrivastava:


dmitsh (approved):

pkg/providers/baremetal/mnnvl.go
```diff
@@ -198,17 +198,26 @@ func getClusterOutput(ctx context.Context, domainMap map[string]domain, nodes []
...
+	clusterUUID := ""
+	domainName := ""
 	for scanner.Scan() {
 		nodeLine := scanner.Text()
 		arr := strings.Split(nodeLine, ":")
 		nodeName := arr[0]
-		clusterUUID := strings.TrimSpace(arr[2])
-		if !domainIDExists(clusterUUID, domainMap) {
-			domainMap[clusterUUID] = domain{
+		itemName := strings.TrimSpace(arr[1])
```
pkedy, line 208: @ritikasrivastava Would it be possible to break out the scanning so that you can write a unit test for this? There is just enough here I think to warrant it.

ritikasrivastava: Sure Phil, Unit tests would come in as a separate [PR](https://jirasw.nvidia.com/browse/NGCC-21892).

PR #53, opened 2025-01-15
simplify mocking payloads

dmitsh:
auto-generate instanceName:nodeName map from the model

### DIFF
--- a/pkg/models/model.go
+++ b/pkg/models/model.go
@@ -19,6 +19,7 @@ package models
 import (
 	"fmt"
 	"os"
+	"sort"
 
 	"gopkg.in/yaml.v3"
 
@@ -30,7 +31,8 @@ type Model struct {
 	CapacityBlocks []*CapacityBlock `yaml:"capacity_blocks"`
 
 	// derived
-	Nodes map[string]*Node
+	Nodes     map[string]*Node
+	Instances []topology.ComputeInstances
 }
 
 type Switch struct {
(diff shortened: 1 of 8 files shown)


pkg/models/model.go
```diff
@@ -123,9 +131,33 @@ func (m *Model) setNodeMap() error {
...
+
+			region, ok := metadata["region"]
+			if !ok {
+				region = "none"
+			}
+			if _, ok := regions[region]; !ok {
+				regions[region] = make(map[string]string)
+
+			}
+			regions[region][name] = name
```
pkedy: Super Nit: "Upserting" into a map usually looks a little more like this. Slightly more efficient but +1 line.

```suggestion
			r, ok := regions[region]
			if !ok {
			    r = make(map[string]string)
			    regions[region] = r
			}
			r[name] = name
```

pkg/models/model.go
```diff
@@ -123,9 +131,33 @@ func (m *Model) setNodeMap() error {
...
 		}
 	}
 
+	regionNames := make([]string, 0, len(regions))
+	for region := range regions {
+		regionNames = append(regionNames, region)
+	}
```
pkedy, line 151: Nit: You could you use `copy` here instead, right?

```suggestion
	regionNames := make([]string, len(regions))
	copy(regionNames, regions)
```

pkedy (approved):
Just a few nits. LGTM.

pkg/models/model.go
dmitsh: regions is a map, regionNames is an array

PR #54, opened 2025-01-16
updated map2string to be generic

kevin-kho:
- Removed `mapOfAny2string`
- Updated `map2string` to be generic. It accepts `map[string]string` or `map[string]any`
- `%v` will work for either map. Using `%v` for a `string` will default to `%s`


pkedy (approved):
LGTM. Good use of generics.

PR #60, opened 2025-01-23
Add test cases

ritikasrivastava:


dmitsh (commented):
Do we use functions like GetBlockWithIBTestSet() anywhere except tests?
Perhaps we can move them from output.go to output_test.go?

ritikasrivastava:
> Do we use functions like GetBlockWithIBTestSet() anywhere except tests? Perhaps we can move them from output.go to output_test.go?

Moved the unexpected functions to _test.go

dmitsh (approved):

PR #72, opened 2025-02-19
Scan the node list instead of partitions for mnnvl

ritikasrivastava:


internal/exec/exec.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 func Exec(ctx context.Context, exe string, args []string, env map[string]string) (*bytes.Buffer, error) {
-	klog.Infof("Execute command %s", strings.Join(append([]string{exe}, args...), " "))
+	klog.V(2).Infof("Execute command %s", strings.Join(append([]string{exe}, args...), " "))
 	cmd := exec.CommandContext(ctx, exe, args...)
```
ritikasrivastava, line 32: It is in the exec.go itself.

PR #119, opened 2025-06-04
make k8s topology node label names configurable

dmitsh:


cmd/topograph/main.go
```diff
@@ -27,15 +27,21 @@ import (
...
 )
 
 var GitTag string
 
 func main() {
 	var cfg string
+	var labelAccelerator, labelBlock, labelSpine, labelDatacenter string
 	var version bool
 	flag.StringVar(&cfg, "c", "/etc/topograph/topograph-config.yaml", "config file")
+	flag.StringVar(&labelAccelerator, "label_acc", k8s.DefaultLabelAccelerator, "K8s node label for accelerated network type")
```
sanjaychatterjee: Can we name the flag `k8s-topology-key-accelerator` instead of `label_acc`? Similarly for others as well.

dmitsh: done

PR #241, opened 2026-04-07
feat: Slinky Engine - Dynamic Nodes Reconciliation

ravisoundar:


pkg/engines/slinky/engine.go
```diff
@@ -321,3 +349,120 @@ func (eng *SlinkyEngine) getPartitionNodes(ctx context.Context, partition string
...
+
+	// Get the topology desiredSpec for the node based on the desired topologies
+	desiredSpec, httpErr := nt.GetNodeTopologySpec(slurmName, topologies)
+	if httpErr != nil {
+		return httperr.NewError(http.StatusBadGateway, fmt.Sprintf("failed to get topology spec for node %s: %v", slurmName, httpErr))
+	}
+
+	//If the topology spec is empty, no topology information is available for the node from the provider.
+	//In this case, we can skip the annotation update to avoid unnecessary node reconfiguration by Slinky.
+	if desiredSpec == "" {
```
bugra-gedik-nv, line 426: What if there is an existing annotation on the node, don't we need to clear it?

pkg/topology/topology.go
```diff
@@ -50,6 +50,10 @@ const (
 	KeyConfigMapPlugin            = "topograph.nvidia.com/plugin"
 	KeyConfigMapBlockSizes        = "topograph.nvidia.com/block-sizes"
 	KeyConfigMapNamespace         = "topograph.nvidia.com/slurm-namespace"
+
+	//Slinky specific annotations and labels
+	KeySlinkyTopologySpec = "topology.slinky.slurm.net/spec"
+	KeySlurmNodeName      = "slurm.node.name"
```
bugra-gedik-nv, line 56: This is a lepton specific annotation, but if I understand the code correctly, if this does not exists (regular slinky) you use the hostname field form the pod spec, which is correct

pkg/translate/tree.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 // toTreeTopology generates SLURM cluster topology config in "topology/tree" format
-func (nt *NetworkTopology) toTreeTopology(wr io.Writer) *httperr.Error {
+func (nt *NetworkTopology) toTreeTopology(wr io.Writer, skeletonOnly bool) *httperr.Error {
```
bugra-gedik-nv, line 20: does this only skip nodes?

I think we only need the top level switches and not the intermediaries.

pkg/engines/slinky/engine_test.go
```diff
@@ -308,3 +320,254 @@ func TestConfigMapAnnotationsAndMetadata(t *testing.T) {
...
+							Status: corev1.ConditionTrue,
+						},
+					},
+				},
+			}
+			_, err = client.CoreV1().Pods("test-ns").Create(context.Background(), pod1, metav1.CreateOptions{})
+			require.NoError(t, err)
+		}
+		// Add config map
+		cm := &corev1.ConfigMap{
```
faganihajizada: All success tests pre-create a ConfigMap; no test covers the `errors.IsNotFound` path.

pkg/translate/yaml.go
```diff
@@ -47,8 +49,37 @@ type Block struct {
 	Nodes string `yaml:"nodes,omitempty"`
 }
 
-// toYamlTopology generates SLURM cluster topology config in YAML format
-func (nt *NetworkTopology) toYamlTopology(wr io.Writer) *httperr.Error {
+func (nt *NetworkTopology) copySkeleton(topologies []*TopologyUnit) []*TopologyUnit {
```
faganihajizada: copySkeleton is a method on **NetworkTopology** but doesn't use the receiver.

pkg/translate/tree.go
ravisoundar: Yes, it only skip nodes for now. I left the intermediate switches as a precaution as I had the below question about it:

Just the top level switches make sense when topograph directly calls `scontrol update` on slurm. In that case, topology update will have the format `Topology=<topo-name>:<root-sw>,<spine-sw>,<leaf-sw>`. From this config, slum will be able to learn about the intermediate switch names, and the mapping between the  spine and root switches.

Where as, when we use slinky topology spec annotation: the spec has the following format: `topology.slinky.slurm.net/spec: <topo-name-01>:<switch-name>`. Where the switch name is the leaf switch name, and the mapping between the spine and root switches will not be available either through topology spec annotation or through topology.yaml. How slurm will know that mapping if we only include the top level switch in the topology.yaml ?

pkg/translate/yaml.go
ravisoundar: Made it as a separate function.

pkg/engines/slinky/engine_test.go
ravisoundar: Updated the some of the success test cases to not pre-create the config map.

pkg/translate/tree.go
bugra-gedik-nv: You are right. I did not notice this. So looks like we'll need reconfigures when new switches are added

pkg/engines/slinky/engine.go
ravisoundar: The desiredSpec would be empty when the provider apis do not know about this `slurmNode`.

One possible scenario I could think of is: the node just joined (i.e. after the API call to the provider but before the k8s NodeList query), so it reached here. In that case, node observer would trigger a reconciliation for the new node, and the node annotation will get updated as part of that. 

Alternatively, if the provider APIs does not know about it because the node is now removed from the topology, then the `slurmd pod and k8s node` would be in the process of getting removed. So, probably updating it doesn't matter. So, the design doc mentioned a no-op in case of a node removal.  

Is there any specific case that would require updating clearing the existing value on the annotation?

pkg/topology/topology.go
ravisoundar: yes. I just created the constant, otherwise, it was the existing logic.

PR #242, opened 2026-04-13
feat(k8s): node observer to retry failed requests after delay

dmitsh:


pkg/node_observer/status_informer.go
```diff
@@ -188,7 +191,9 @@ func (s *StatusInformer) processEvent() bool {
 	_, err := httpreq.DoRequestWithRetries(s.reqFunc, false)
 	if err != nil {
 		klog.Errorf("failed to send HTTP request: %v", err)
+		s.queue.AddAfter(item, 5*time.Minute)
```
bugra-gedik-nv: Is there any concern about the queue filling up?

PR #259, opened 2026-04-17
docs(k8s): add Kubernetes access patterns section

resker:
## Summary

Adds a new \"Exposing the Topograph API\" section to `docs/engines/k8s.md` that documents how the API is accessed in Kubernetes deployments and makes the authentication story explicit.

## What it covers

- Access pattern matrix: ClusterIP + port-forward (default), in-cluster callers, NodePort/LoadBalancer, traditional Ingress. Notes Gateway API is not yet supported.
- Default ClusterIP behavior: the chart does not expose the API externally by default; in-cluster components use the Service DNS name; operators debug via port-forward.
- Ingress configuration example with common auth patterns (oauth2-proxy, Istio RequestAuthentication, mTLS termination).
- `/metrics` endpoint and ServiceMonitor for Prometheus Operator scraping.
- NetworkPolicy starter for cluster-internal traffic control (chart does not yet ship a bundled template).

## Why

The current chart supports a mix of access patterns via `global.service.type`, `ingress.enabled`, and `serviceMonitor.enabled`, but the access/auth story is scattered across `values.yaml` and `NOTES.txt`. This consolidates the operator-facing guidance into one section.

## Test plan

- [x] Markdown renders cleanly (checked with GitHub preview)
- [x] All references to chart values verified against `charts/topograph/values.yaml` and templates
- [x] NetworkPolicy example `apiVersion` and selectors are valid against K8s 1.24+
- [x] Service DNS name format matches the default `Service` template in `charts/topograph/templates/service.yaml`

### DIFF
--- a/docs/engines/k8s.md
+++ b/docs/engines/k8s.md
@@ -141,9 +141,11 @@ By default, `global.service.type: ClusterIP` and `ingress.enabled: false`. This 
 - Cluster operators can reach the API via port-forward for debugging:
 
 ```bash
-kubectl -n <namespace> port-forward svc/topograph 49021:49021
+kubectl -n <namespace> port-forward svc/<release>-topograph 49021:49021
 curl http://localhost:49021/healthz
 ```
+
+The Service name is `<release>-topograph` by default (rendered from the chart's `fullname` template); substitute your Helm release name and the namespace you deployed it into.
 
 This is the recommended pattern for production deployments where Topograph is consumed only by in-cluster callers.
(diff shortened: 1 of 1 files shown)


resker:
Tracking issue: #261 — covers the full roadmap (access-patterns docs → bundled NetworkPolicy template → Gateway API support) and flags the larger authentication/attestation work area as separate future scope.

dmitsh (approved):

PR #268, opened 2026-04-18
chore(github): update fern github actions

dmitsh:


resker:
Small typo in `fern-docs-ci.yml`: the new `paths:` filter references `.github/workflows/fern-docs-ci.yaml`, but the actual file is `fern-docs-ci.yml` (one `l`). Looks like it was copied from AICR's workflow (where the file is `.yaml`).

The filter as written won't match changes to the workflow file itself. Two ways to fix:

- Change to `.yml` to match the actual filename, or
- Drop the path filter entirely — NVSentinel's equivalent has no workflow-file filter and this PR otherwise aligns cleanly to NVSentinel's pattern.

PR #320, opened 2026-05-14
fix(slinky): tighten pods exec RBAC for partition discovery

dmitsh:
## Description
Grant pods/exec only when Slinky may use legacy partition discovery.
Treat explicit empty node lists as resolved so they do not fall back to
scontrol, keeping runtime behavior aligned with rendered RBAC.

## Checklist


ravisoundar (approved):
lgtm

PR #341, opened 2026-05-24
fix(k8s): prefer GPU clique label for accelerator domains

dmitsh:
## Description
Use nvidia.com/gpu.clique as the authoritative accelerator-domain
signal when it is already present on Kubernetes nodes. In that case,
skip writing the Topograph accelerator label while still applying
leaf, spine, and core topology labels.

For infiniband-k8s, add an optional clusterIDLabel provider parameter
so deployments can reuse the GPU Operator clique label instead of
collecting the same value through nvidia-smi.

## Checklist


ravisoundar (dismissed):
lgtm

PR #348, opened 2026-06-12
fix(server): preserve replacement timer in trailing delay queue

dmitsh:
## Description
Prevent a running stale timer callback from deleting a newer timer for the
same request hash. Add a regression test covering the replacement-timer race.

## Checklist


ravisoundar (approved):
lgtm

PR #350, opened 2026-06-13
fix(server): include nodes and creds in request dedupe

fallintoplace:
## Description
Fixes request aggregation collisions where requests with the same provider and engine params but different explicit node mappings or credentials could share the same request ID.

The request hash now includes canonicalized node mappings and an HMAC digest of the effective credential map. The server resolves configured credentials before enqueue and lookup, so config-backed requests use the same key for both endpoints. `TrailingDelayQueue.Get` also returns a snapshot of cached completion state so polling cannot race with completion updates.

Validation:
- `make qualify LINTER_BIN=/Users/hoangvu/go/bin/golangci-lint`

## Checklist

### DIFF
--- a/pkg/server/http_server.go
+++ b/pkg/server/http_server.go
@@ -204,6 +204,8 @@ func readRequest(w http.ResponseWriter, r *http.Request) *topology.Request {
 		}
 	}
 
+	tr.Provider.Creds = checkCredentials(tr.Provider.Creds, srv.cfg.Credentials)
+
 	klog.Info(tr.String())
 
 	if err = validate(tr); err != nil {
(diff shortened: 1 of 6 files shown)


dmitsh:
Hi @fallintoplace ,
Thank you for your contribution.
I understand the rationale for including credentials and the node set in the request hash. However, we didn't identify a use case that requires this, and for the sake of simplicity, we decided not to include them.
Do you have a specific use case where this is actually required?
In the meantime, could you please submit a separate PR with your changes to pkg/server/trailing_delay_queue.go ?
Those changes fix a race condition and would be valuable independently.

dmitsh:
Closing the PR for now, as nodes and credentials are not currently planned to be included in the payload hash.

PR #351, opened 2026-06-15
fix(server): snapshot queue completion results

fallintoplace:
## Description
Split out from #350 at maintainer request.

`TrailingDelayQueue.Get` returned the live `*Completion` stored in the queue cache. The queue lock protected the lookup itself, but callers read the returned pointer after the lock was released while the timer callback could still update the same `Completion`. That showed up as a race between polling `/v1/topology` and request completion.

This PR returns a shallow snapshot of the cached completion instead. It also adds a focused regression test that proves an accepted-result snapshot does not mutate after the worker completes.

Validation:
- `go test -race ./pkg/server`
- `make qualify LINTER_BIN=/Users/hoangvu/go/bin/golangci-lint`

## Checklist

### DIFF
--- a/pkg/server/trailing_delay_queue.go
+++ b/pkg/server/trailing_delay_queue.go
@@ -122,7 +122,8 @@ func (q *TrailingDelayQueue) Get(hash string) *Completion {
 	defer q.mutex.Unlock()
 
 	if res, ok := q.store.Get(hash); ok {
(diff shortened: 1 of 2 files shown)


dmitsh (approved):

dmitsh:
@fallintoplace , thank you for your contribution!

PR #360, opened 2026-06-24
chore(helm): default image tag to chart appVersion

giuliocalzo:
## Description

Make every Topograph image reference derive from a single source — the chart `appVersion` — instead of repeating a hardcoded tag in each `values.yaml`. The chart `version`/`appVersion` stay at `0.4.0`, so the rendered images are unchanged (`ghcr.io/nvidia/topograph:v0.4.0`); this is a wiring/refactor change, not a release bump.

- Empty the `image.tag` (and `initc.image.tag`) defaults in each chart's `values.yaml` so they fall back to `.Chart.AppVersion`.
- Add `*.image` / `*.initImage` helper templates so the main chart and both subcharts resolve the image reference (with the `| default .Chart.AppVersion` fallback) in one place.
- Apply the `appVersion` fallback to the `node-data-broker` init container and the `helm test` hook pods (`test-healthz`, `test-metrics`) so they resolve correctly when the tag is empty.
- Make the `node-observer` `wait` init container image configurable via a new `waitImage` value (previously hardcoded `curlimages/curl:8.13.0`).
- Refresh chart goldens under `tests/charts/` (the only rendered delta is the `wait` image string now being quoted).

## Checklist

- [x] `make chart-test` passes
- [x] Documentation impact evaluated — chart values default behavior documented via inline `values.yaml` comments
- [x] No `pkg/topology/` changes
- [x] Every commit has a DCO sign-off

### DIFF
--- a/charts/topograph/Chart.yaml
+++ b/charts/topograph/Chart.yaml
@@ -16,18 +16,18 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
+version: 0.4.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: "v0.4.0"
+appVersion: "v0.4.1"
 
 dependencies:
   - name: node-data-broker
-    version: 0.4.0
+    version: 0.4.1
     repository: "file://charts/node-data-broker"
   - name: node-observer
-    version: 0.4.0
+    version: 0.4.1
     repository: "file://charts/node-observer"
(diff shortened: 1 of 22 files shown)


charts/topograph/charts/node-data-broker/Chart.yaml
```diff
@@ -16,7 +16,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
+version: 0.4.1
```
dmitsh: let's keep 0.4.0

charts/topograph/charts/node-observer/Chart.yaml
```diff
@@ -16,7 +16,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.4.0
+version: 0.4.1
```
dmitsh: let's keep 0.4.0

charts/topograph/Chart.yaml
```diff
@@ -26,8 +26,8 @@ appVersion: "v0.4.0"
 
 dependencies:
   - name: node-data-broker
-    version: 0.4.0
+    version: 0.4.1
```
dmitsh: let's keep 0.4.0

ravisoundar (approved):
lgtm

PR #362, opened 2026-06-25
feat(engine/slinky): discover partitions via the controller pod

giuliocalzo:
## Description

A Slinky cluster always runs a controller (`slurmctld`), but login pods are optional. The Slinky engine's legacy partition discovery (`getPartitionNodes`) execed `scontrol show partition` only in a pod labeled `app.kubernetes.io/component: login`, so discovery failed on clusters installed without login pods.

This changes discovery to prefer the **controller** pod and fall back to a **login** pod when one is present:

- `pkg/engines/slinky/engine.go` — iterate `["controller", "login"]` component labels; exec in the first running pod found. Component label key/values extracted into named constants. The not-found error now names both components and the namespace.
- `pkg/engines/slinky/engine_test.go` — new `TestGetPartitionNodes` covers the dynamic-nodes short-circuit, both parameter-validation errors, and the no-running-pods path (the `ExecInPod` success path isn't exercisable with a fake clientset).
- `charts/topograph/templates/rbac.yaml` — `pods/exec` rationale comment updated to reflect controller-first discovery.
- `docs/engines/slinky.md` — partition-discovery fallback row updated; the controller is always present, login pods are optional.

This is backward compatible: clusters that only have login pods (and no controller match) still work via the fallback.

## Checklist


dmitsh (dismissed):

giuliocalzo:
Addressed the P1 too: `ExecInPod` failures are now non-fatal per component, so a failed exec on the controller falls back to the login pod (symmetric with the listing behavior) instead of returning early. Added tests for the controller-exec-fails→login and all-components-fail cases.

giuliocalzo:
Pushed `1c5ba07` with the P1 exec-fallback fix on top of the latest `main` merge. Both concerns are now addressed: listing errors and exec errors are non-fatal per component, so discovery falls back controller → login instead of aborting. Tests cover the controller-exec-fails→login and all-components-fail cases. Ready for another look, @dmitsh.

dmitsh (approved):

PR #363, opened 2026-06-25
feat(node-observer): wait for topograph health in-process

giuliocalzo:
## Description

Replace the node-observer chart's curl-based `wait` init container with an in-process health wait in the `node-observer` binary, removing the dependency on the `curlimages/curl` image for this subchart.

- **node-observer** waits for the topograph API to become healthy before starting its watch loop. The controller polls `/healthz` (derived from `generateTopologyUrl`) every 2s and gives up after 1m, returning an error so the pod restarts if topograph never comes up. Timeout errors report the actual elapsed time rather than the nominal deadline.
- The `wait` init container, the `waitImage` value, and the `node-observer.waitImage` helper are removed.
- Helm-unittest coverage and snapshots updated for the slimmer Deployment.

Related follow-up (separate PR): node-data-broker main-container / health-endpoint changes.

## Checklist

## Test plan
- [x] `go test ./pkg/node_observer/...`
- [x] `make chart-test`

### DIFF
--- a/charts/topograph/charts/node-observer/templates/_helpers.tpl
+++ b/charts/topograph/charts/node-observer/templates/_helpers.tpl
@@ -38,13 +38,6 @@ Container image reference. The tag defaults to the chart appVersion when unset.
 {{- end }}
 
 {{/*
-Wait init container image reference.
-*/}}
-{{- define "node-observer.waitImage" -}}
-{{- printf "%s:%s" .Values.waitImage.repository .Values.waitImage.tag -}}
-{{- end }}
-
-{{/*
 Common labels
 */}}
 {{- define "node-observer.labels" -}}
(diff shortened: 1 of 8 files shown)


dmitsh:
@giuliocalzo , could you split it into 2 PRs, one for node-observer, one for node-data-broker?
Could you also open an issue for each PR?

giuliocalzo:
Closing in favor of #367.

Both PRs remove the node-observer `wait` init container and move readiness handling into the main process, but they touch the same chart templates and controller wiring:

- **#363** — polls `/healthz` at startup, then starts the existing watch loop. Does not re-trigger if the Topograph API restarts later.
- **#367** — watches the Topograph API pod (Ready + restart count) and triggers regeneration after API restarts (closes #366).

#367 is the broader maintainer fix for the same problem space, so this PR is superseded. The node-data-broker work continues separately in #368.

Issue #263, opened 2026-04-17
Add Chainsaw conformance tests for the Kubernetes engine


resker:
## Problem

`docs/engines/k8s.md` currently ends with:

> ## Validation and Testing
> TBD

Topograph has Go unit tests and integration fixtures (`tests/integration/payloads/*.json`), but no end-to-end cluster-level suite that asserts "given a deployed chart, do the right labels and annotations actually land on nodes?" This gap is felt most acutely when changing the Helm chart, Node Observer, or engine output — there's no quick way to verify the full deployment-to-labels flow without manual cluster testing.

## Proposed approach

Adopt **[Chainsaw](https://kyverno.github.io/chainsaw/)** (Kyverno team, Apache 2.0) — the declarative Kubernetes E2E test framework used widely across Kubernetes OSS. Tests are YAML that run `apply → wait → assert → cleanup` against a real cluster (kind in CI). Assertions use JMESPath-style expressions to verify cluster state, which fits Topograph's output model precisely: node labels, annotations, and ConfigMaps are exactly what Chainsaw is designed to assert.

**Why Chainsaw specifically:**

1. Topograph's Kubernetes engine writes cluster state — Chainsaw is purpose-built for asserting cluster state
2. The `test` provider + toposim models give deterministic inputs (no real hardware, no cloud credentials) — tests stay reproducible
3. Node Observer reactivity is testable — Chainsaw can add/remove nodes and assert the re-labeling occurs within the aggregation delay
4. Strong CNCF ecosystem alignment — Chainsaw is the common pattern for declarative Kubernetes conformance testing
5. Precedent in NVIDIA's own OSS — [NVIDIA/aicr](https://github.com/NVIDIA/aicr) uses Chainsaw for AI Conformance testing (`tests/chainsaw/ai-conformance/common/assert-kai-scheduler.yaml` etc.), which gives us a known-good structural template

## Self-contained local development

Chainsaw + the `test` provider + toposim together enable a self-contained local development loop — contributors can clone the repo, spin up kind, install the chart, and run the full E2E suite on their own workstation without NVIDIA hardware, cloud provider credentials, or access to a shared cluster. This covers the majority of engine-facing code paths:

| Code path | Self-contained with Chainsaw? |
|---|---|
| API server, aggregation, validation | ✅ |
| k8s engine output (labels) | ✅ |
| Slinky engine output (ConfigMap) | ✅ |
| Node Observer reactivity to node add/remove events | ✅ |
[...]

PR #365, opened 2026-06-26
test(e2e): e2e k8s environment tests using chainsaw

ravisoundar:
## Description
End-2-end tests for the Kubernetes environment using Chainsaw.
Addresses issue #263 

## Checklist


tests/chainsaw/k8s/label-application/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+  requestAggregationDelay: 1s
+
+node-observer:
+  topograph:
+    trigger:
+      nodeSelector:
+        kubernetes.io/os: linux
+
+node-data-broker:
+  enabled: false
```
ArangoGutierrez, line 19: Every suite's values.yaml disables the node-data-broker DaemonSet but leaves `node-observer.topograph.nodeDataBroker.enabled` at its default `true`, so the observer's `allBrokerPodsReady()` waits forever for broker pods that never exist and `/v1/generate` is never triggered - all six suites should time out on their assert steps. Per the node-data-broker subchart's values comment, also set `node-observer.topograph.nodeDataBroker.enabled: false` (same fix in the other five values.yaml files).

tests/chainsaw/.chainsaw.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  timeouts:
+    apply: 120s
+    assert: 90s
+    cleanup: 60s
+    delete: 60s
+    error: 30s
+    exec: 120s
+  fullName: true
+  parallel: 1
```
ArangoGutierrez, line 14: Chainsaw only auto-discovers a file named `.chainsaw.yaml`, and neither the `e2e`/`e2e-local` Makefile targets nor the workflow pass `--config`, so this file - including `parallel: 1` - is silently ignored. With default parallelism the six suites race: all use the fixed release name `topograph`, so the release-scoped ClusterRole/ClusterRoleBinding names collide and each suite's pre-install `kubectl delete clusterrole` tears down RBAC another suite owns. Renaming to `.chainsaw.yaml` or passing `--config` fixes it.

Makefile
```diff
@@ -115,6 +115,49 @@ docker-buildx:
...
+
+# Build the image, create a 4-worker kind cluster, load the image, run all
+# Chainsaw suites, and destroy the cluster.  Requires kind and chainsaw.
+.PHONY: e2e-local
+e2e-local: chainsaw-install image-build
+	kind create cluster --name $(KIND_CLUSTER) \
+	  --config tests/chainsaw/kind-config.yaml --wait 120s \
+	  || kind get clusters | grep -q "^$(KIND_CLUSTER)$$"
+	kind load docker-image $(IMAGE_REPO):$(E2E_IMAGE_TAG) --name $(KIND_CLUSTER)
+	KUBECONFIG="$$(kind get kubeconfig --name $(KIND_CLUSTER))" \
```
ArangoGutierrez: `kind get kubeconfig` prints kubeconfig content to stdout, not a path, so this sets `KUBECONFIG` to inline YAML and every kubectl/helm/chainsaw call in `e2e-local` fails to find a config. Write it to a temp file (or use `kind export kubeconfig`) and point `KUBECONFIG` at that.

ArangoGutierrez (commented):
Found 5 issues.

1. Every suite's values.yaml disables the node-data-broker DaemonSet but leaves `node-observer.topograph.nodeDataBroker.enabled` at its default `true`, so the observer's `allBrokerPodsReady()` waits forever for broker pods that never exist and `/v1/generate` is never triggered - all six suites should time out on their assert steps. Per the node-data-broker subchart's values comment, also set `node-observer.topograph.nodeDataBroker.enabled: false` (same fix in the other five values.yaml files). (tests/chainsaw/k8s/label-application/values.yaml:19)
2. Chainsaw only auto-discovers a file named `.chainsaw.yaml`, and neither the `e2e`/`e2e-local` Makefile targets nor the workflow pass `--config`, so this file - including `parallel: 1` - is silently ignored. With default parallelism the six suites race: all use the fixed release name `topograph`, so the release-scoped ClusterRole/ClusterRoleBinding names collide and each suite's pre-install `kubectl delete clusterrole` tears down RBAC another suite owns. Renaming to `.chainsaw.yaml` or passing `--config` fixes it. (tests/chainsaw/chainsaw-config.yaml:14)
3. `kind get kubeconfig` prints kubeconfig content to stdout, not a path, so this sets `KUBECONFIG` to inline YAML and every kubectl/helm/chainsaw call in `e2e-local` fails to find a config. Write it to a temp file (or use `kind export kubeconfig`) and point `KUBECONFIG` at that. (Makefile:152)
4. go.mod requires Go 1.25.11, and the four pins in go.yml were deliberately bumped from 1.25.9 to 1.25.11 (cb0362e) to clear reachable stdlib vulnerabilities - this reintroduces the pre-fix toolchain. Pin `1.25.11` here too. (.github/workflows/e2e.yml:32)
5. This row claims `label-application` verifies `accelerator` labels, but the suite's topology model has no NVLink data and its assert step only checks `leaf` and `spine` (tests/chainsaw/README.md lists it correctly). The table also omits the `slinky/block-complement` and `slinky/dynamic-nodes` suites shipped in this PR. (docs/engines/k8s.md:357)

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,100 @@
...
+    permissions:
+      contents: read
+
+    steps:
+    - uses: actions/checkout@v5
+
+    - name: Set up Go
+      uses: actions/setup-go@v6
+      with:
+        go-version: '1.25.9'
```
ArangoGutierrez: go.mod requires Go 1.25.11, and the four pins in go.yml were deliberately bumped from 1.25.9 to 1.25.11 (cb0362e) to clear reachable stdlib vulnerabilities - this reintroduces the pre-fix toolchain. Pin `1.25.11` here too.

docs/engines/k8s.md
```diff
@@ -344,6 +344,52 @@ tests:
...
+
+`helm test` verifies that a deployed instance is healthy. To verify that the engine actually **applies correct topology labels** to nodes, use the Chainsaw E2E suite in `tests/chainsaw/`.
+
+[Chainsaw](https://kyverno.github.io/chainsaw/) is Kyverno's declarative E2E framework. Each suite drives `apply → wait → assert → cleanup` against a real cluster using the built-in **test provider** — no cloud credentials required.
+
+#### Test suites
+
+| Suite | What it checks |
+|---|---|
+| `k8s/label-application` | `leaf`, `spine`, and `accelerator` labels applied to nodes after generation |
```
ArangoGutierrez: This row claims `label-application` verifies `accelerator` labels, but the suite's topology model has no NVLink data and its assert step only checks `leaf` and `spine` (tests/chainsaw/README.md lists it correctly). The table also omits the `slinky/block-complement` and `slinky/dynamic-nodes` suites shipped in this PR.

tests/chainsaw/k8s/label-application/values.yaml
ravisoundar: With the new chart changes `node-observer.topograph.nodeDataBroker.enabled` is not required.

tests/chainsaw/.chainsaw.yaml
ravisoundar: Renamed it to .chainsaw.yaml now

Makefile
ravisoundar: Fixed it now.

.github/workflows/e2e.yml
ravisoundar: Fixed it now.

docs/engines/k8s.md
ravisoundar: Fixed it now.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,100 @@
...
+    permissions:
+      contents: read
+
+    steps:
+    - uses: actions/checkout@v5
+
+    - name: Set up Go
+      uses: actions/setup-go@v6
+      with:
+        go-version: '1.25.11'
```
ArangoGutierrez, line 32: This pin went stale mid-review: #395 has since bumped the repo to Go 1.26.5 (go.mod and all four pins in go.yml). On runners where toolchain auto-fetch is restricted this fails outright; everywhere else it silently downloads 1.26.5 anyway, so best to align it. My earlier 1.25.11 suggestion predates the #395 bump.

tests/chainsaw/k8s/label-application/chainsaw-test.yaml
```diff
@@ -0,0 +1,155 @@
...
+    - script:
+        timeout: 120s
+        content: |
+          set -euo pipefail
+          REPO_ROOT=$(git rev-parse --show-toplevel)
+          IMAGE_REPO="${TOPOGRAPH_IMAGE_REPO:-ghcr.io/nvidia/topograph}"
+          IMAGE_TAG="${TOPOGRAPH_IMAGE_TAG:-}"
+          PULL_POLICY="${TOPOGRAPH_IMAGE_PULL_POLICY:-IfNotPresent}"
+
+          kubectl delete clusterrole topograph topograph-node-observer \
```
ArangoGutierrez, line 80: This delete (and the same pattern in all six suites, setup and cleanup) removes cluster-scoped objects that share names with a real Topograph deployment's RBAC, against whatever cluster KUBECONFIG currently points at. Since the README documents running `make e2e` against an arbitrary context, a stale KUBECONFIG would silently strip a live cluster's Topograph RBAC - consider guarding on cluster identity, e.g. only proceed when the current context is the expected kind cluster.

Makefile
```diff
@@ -115,6 +115,52 @@ docker-buildx:
...
+	$(CHAINSAW_BIN) test --test-dir tests/chainsaw
+
+# Build the image, create a 4-worker kind cluster, load the image, run all
+# Chainsaw suites, and destroy the cluster.  Requires kind and chainsaw.
+.PHONY: e2e-local
+e2e-local: chainsaw-install image-build
+	kind create cluster --name $(KIND_CLUSTER) \
+	  --config tests/chainsaw/kind-config.yaml --wait 120s \
+	  || kind get clusters | grep -q "^$(KIND_CLUSTER)$$"
+	kind load docker-image $(IMAGE_REPO):$(E2E_IMAGE_TAG) --name $(KIND_CLUSTER)
```
ArangoGutierrez, line 151: This `kind load` runs as its own recipe line, outside the shell block that captures `E2E_STATUS` and deletes the cluster, so a load failure aborts the target and orphans the just-created kind cluster. Folding it into the guarded block keeps the teardown promise.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Install kind
+      run: go install sigs.k8s.io/kind@latest
```
ArangoGutierrez, line 35: `kind@latest` floats, so a new kind release can change cluster behavior between runs with no way to pin or bisect. Worth the same treatment as the chainsaw install below: a pinned, overridable version.

Makefile
```diff
@@ -115,6 +115,52 @@ docker-buildx:
...
+
+# Load the locally-built image into an existing kind cluster with the correct
+# E2E_IMAGE_TAG.  Use this before running make e2e against a local kind cluster:
+#   make kind-load KIND_CLUSTER=topograph-test && make e2e
+.PHONY: kind-load
+kind-load:
+	kind load docker-image $(IMAGE_REPO):$(E2E_IMAGE_TAG) --name $(KIND_CLUSTER)
+
+# Run all Chainsaw E2E suites against the current KUBECONFIG context.
+# For a pre-pushed registry image: set TOPOGRAPH_IMAGE_REPO and TOPOGRAPH_IMAGE_TAG.
```
ArangoGutierrez, line 136: This comment points users at `TOPOGRAPH_IMAGE_REPO`/`TOPOGRAPH_IMAGE_TAG`, but the recipe unconditionally overwrites both from `$(IMAGE_REPO)`/`$(E2E_IMAGE_TAG)`, and the new README explicitly warns against setting them as env vars. Recommending the `IMAGE_REPO`/`E2E_IMAGE_TAG` make overrides here would keep the two in agreement.

AGENTS.md
```diff
@@ -43,6 +43,7 @@ charts/topograph/     # Helm chart for all Kubernetes components; tests/ holds t
 CHANGELOG.md          # Release history (Keep a Changelog format); update [Unreleased] for user-facing PRs
 docs/                 # Public-facing docs — overview.md, architecture.md, api.md + providers/, engines/, reference/ subdirectories
 tests/models/         # YAML simulation fixtures
+tests/chainsaw/       # Chainsaw E2E test suites (label-application, label-truncation, node-observer, slinky)
```
ArangoGutierrez, line 46: The suite list names `node-observer`, but no such suite exists under `tests/chainsaw/` - the actual suites are `k8s/{label-application,label-truncation}` and `slinky/{block-complement,dra-provider,dynamic-nodes,tree-topology}`. The synced line in `.claude/CLAUDE.md` needs the same fix.

ArangoGutierrez (commented):
Found 6 issues.

1. This pin went stale mid-review: #395 has since bumped the repo to Go 1.26.5 (go.mod and all four pins in go.yml). On runners where toolchain auto-fetch is restricted this fails outright; everywhere else it silently downloads 1.26.5 anyway, so best to align it. My earlier 1.25.11 suggestion predates the #395 bump. (.github/workflows/e2e.yml:32)
2. This delete (and the same pattern in all six suites, setup and cleanup) removes cluster-scoped objects that share names with a real Topograph deployment's RBAC, against whatever cluster KUBECONFIG currently points at. Since the README documents running `make e2e` against an arbitrary context, a stale KUBECONFIG would silently strip a live cluster's Topograph RBAC - consider guarding on cluster identity, e.g. only proceed when the current context is the expected kind cluster. (tests/chainsaw/k8s/label-application/chainsaw-test.yaml:80)
3. This `kind load` runs as its own recipe line, outside the shell block that captures `E2E_STATUS` and deletes the cluster, so a load failure aborts the target and orphans the just-created kind cluster. Folding it into the guarded block keeps the teardown promise. (Makefile:151)
4. `kind@latest` floats, so a new kind release can change cluster behavior between runs with no way to pin or bisect. Worth the same treatment as the chainsaw install below: a pinned, overridable version. (.github/workflows/e2e.yml:35)
5. This comment points users at `TOPOGRAPH_IMAGE_REPO`/`TOPOGRAPH_IMAGE_TAG`, but the recipe unconditionally overwrites both from `$(IMAGE_REPO)`/`$(E2E_IMAGE_TAG)`, and the new README explicitly warns against setting them as env vars. Recommending the `IMAGE_REPO`/`E2E_IMAGE_TAG` make overrides here would keep the two in agreement. (Makefile:136)
6. The suite list names `node-observer`, but no such suite exists under `tests/chainsaw/` - the actual suites are `k8s/{label-application,label-truncation}` and `slinky/{block-complement,dra-provider,dynamic-nodes,tree-topology}`. The synced line in `.claude/CLAUDE.md` needs the same fix. (AGENTS.md:46)

PR #379, opened 2026-07-02
feat(engine/slinky): report why useGpuCliqueLabel found no matching nodes

giuliocalzo:
## Description

When the Slinky engine runs with `useGpuCliqueLabel=true` and cannot build any block domains, it previously returned a generic error:

```
useGpuCliqueLabel=true but no matching nodes found; check label "nvidia.com/gpu.clique" and annotation "topograph.nvidia.com/instance"
```

This gave operators no way to tell *which* nodes were examined or *why* each was skipped. In particular, the `topograph.nvidia.com/instance` annotation is written per-node by the node-data-broker DaemonSet, so a node with the clique label but no annotation points at a broker that hasn't annotated that specific node yet — but the old message hid this.

This PR makes the failure actionable:

- The `502` error now reports how many nodes were scanned and a breakdown of why each was skipped: no Slurm node mapping (no Ready slurmd pod), missing `nvidia.com/gpu.clique` label, or missing the `topograph.nvidia.com/instance` annotation.
- It lists the node names that carry the clique label but are missing the broker-written annotation (capped to keep the message bounded on large clusters).
- The per-node warning now includes the node name, its Slurm name, and the clique value.

Example new message:

```
useGpuCliqueLabel=true but no matching nodes found; check label "nvidia.com/gpu.clique" and annotation "topograph.nvidia.com/instance". Scanned 8 node(s): 2 without a SLURM node mapping (no Ready slurmd pod), 3 missing the "nvidia.com/gpu.clique" label, 3 with the label but missing the "topograph.nvidia.com/instance" annotation (nodes missing annotation: gpu-node-1, gpu-node-4, gpu-node-7)
```

## Checklist

### DIFF
--- a/pkg/engines/slinky/engine.go
+++ b/pkg/engines/slinky/engine.go
@@ -261,6 +261,13 @@ func getComputeInstances(nodes *corev1.NodeList, nodeMap map[string]string) ([]t
 }
 
 func withGPUCliqueDomains(graph *topology.Graph, clusterNodes *clusterNodes) (*topology.Graph, *httperr.Error) {
+	// No nodes selected at all is a distinct failure from "nodes exist but none
+	// matched": it usually points at a too-narrow engine nodeSelector.
+	if len(clusterNodes.nodes.Items) == 0 {
+		return nil, httperr.NewError(http.StatusBadGateway,
+			"no selected Kubernetes nodes found; check engine nodeSelector")
+	}
+
 	domains := topology.NewDomainMap()
 
 	// Diagnostic counters to explain why no domains were built. The instance
--- a/pkg/engines/slinky/engine.go
+++ b/pkg/engines/slinky/engine.go
@@ -262,21 +262,35 @@ func getComputeInstances(nodes *corev1.NodeList, nodeMap map[string]string) ([]t
 
 func withGPUCliqueDomains(graph *topology.Graph, clusterNodes *clusterNodes) (*topology.Graph, *httperr.Error) {
 	domains := topology.NewDomainMap()
+
+	// Diagnostic counters to explain why no domains were built. The instance
+	// annotation is written per-node by the node-data-broker DaemonSet, so a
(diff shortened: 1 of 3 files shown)


pkg/engines/slinky/engine.go
```diff
@@ -262,21 +262,35 @@ func getComputeInstances(nodes *corev1.NodeList, nodeMap map[string]string) ([]t
 
 func withGPUCliqueDomains(graph *topology.Graph, clusterNodes *clusterNodes) (*topology.Graph, *httperr.Error) {
```
dmitsh, line 263: nit: add this check for better debugging:
if len(clusterNodes.nodes.Items) == 0 {
		return nil, httperr.NewError(http.StatusBadGateway,
			"no selected Kubernetes nodes found; check engine nodeSelector")
	}

giuliocalzo: Good call — added in 8468abe. When the engine selects zero Kubernetes nodes, `withGPUCliqueDomains` now returns `no selected Kubernetes nodes found; check engine nodeSelector` (with a dedicated unit test), keeping it distinct from the "nodes exist but none matched" diagnostic. Thanks!

giuliocalzo:
Thanks for the review @dmitsh. Pushed `8468abe` addressing your suggestion:

- `withGPUCliqueDomains` now short-circuits with a dedicated error — `no selected Kubernetes nodes found; check engine nodeSelector` — when zero Kubernetes nodes are selected, keeping it distinct from the "nodes exist but none matched the clique label/annotation" diagnostic.
- Added `TestWithGPUCliqueDomainsNoSelectedNodes` for the empty-selection path; all `TestWithGPUCliqueDomains*` tests pass and the CHANGELOG entry is updated.

The branch was also rebased onto latest `main` earlier to clear the CHANGELOG conflict. Ready for another look when you have a moment.

dmitsh (approved):

PR #382, opened 2026-07-07
docs(design): propose NFD engine for topology groups

dmitsh:
## Checklist

### DIFF
--- a/docs/design/nfd-engine-sdd.md
+++ b/docs/design/nfd-engine-sdd.md
@@ -1,0 +1,221 @@
+# Software Design Document: NFD Engine
+
+## Status
+
+Draft.
+
+## Summary
+
+Add an experimental `nfd` engine that converts Topograph's canonical
+`topology.Graph` into Node Feature Discovery (NFD) `NodeFeatureGroup` objects.
+The engine creates one group for each distinct topology label value, such as one
+group for each accelerator domain, leaf switch, spine switch, and core switch.
+
+This should not replace the current `k8s` engine. The `k8s` engine writes node
+labels that can be consumed by native Kubernetes affinity and topology-aware
+schedulers. The `nfd` engine would publish equivalent group membership through
+NFD CRs for consumers that already watch NFD.
+
+## Background
+
+Topograph already maps topology into four Kubernetes label dimensions:
+
+- `network.topology.nvidia.com/accelerator`
+- `network.topology.nvidia.com/leaf`
+- `network.topology.nvidia.com/spine`
+- `network.topology.nvidia.com/core`
+
+NFD `NodeFeatureGroup` is an alpha NFD API. NFD master watches
+`NodeFeatureGroup` objects, evaluates their feature-group rules, and writes the
+matching nodes into `.status.nodes`. The rules use the same feature-matching
+model as NFD `NodeFeatureRule`. See the NFD
+[custom resources](https://kubernetes-sigs.github.io/node-feature-discovery/master/usage/custom-resources.html#nodefeaturegroup)
+and
+[API reference](https://kubernetes-sigs.github.io/node-feature-discovery/master/reference/generated-nfd-api-reference.html#nodefeaturegroup)
+for the current resource shape.
+
+## Goals
+
+- Add a new engine named `nfd`.
+- Keep provider output unchanged. Providers still return `topology.Graph`.
+- Reuse the k8s engine's graph-to-topology-label mapping.
+- Create `NodeFeatureGroup` objects for distinct topology values.
+
+## Non-Goals
+
+- Do not make NFD required for existing Kubernetes deployments.
+
+## Proposed Behavior
+
+The `nfd` engine should perform three steps:
+
+1. Build the same per-node topology map produced by the `k8s` engine, but do not create node labels.
+2. Publish per-node topology values as NFD `NodeFeature` input.
+3. Create or update one `NodeFeatureGroup` per distinct topology value.
+
+Example generated feature input for `node-a`:
+
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeature
+metadata:
+  name: topograph-node-a
+  labels:
+    nfd.node.kubernetes.io/node-name: node-a
+    app.kubernetes.io/managed-by: topograph
+spec:
+  features:
+    attributes:
+      topograph.network:
+        elements:
+          accelerator: nvl3
+          leaf: leaf-12
+          spine: spine-2
+          core: core-1
+```
+
+Example generated group for one leaf switch:
+
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureGroup
+metadata:
+  name: topograph-leaf-x3f91c4a0
+  labels:
+    app.kubernetes.io/managed-by: topograph
+    topograph.nvidia.com/group-type: leaf
+  annotations:
+    topograph.nvidia.com/label-key: network.topology.nvidia.com/leaf
+    topograph.nvidia.com/label-value: leaf-12
+spec:
+  featureGroupRules:
+    - name: leaf equals leaf-12
+      matchFeatures:
+        - feature: topograph.network
+          matchExpressions:
+            leaf:
+              op: In
+              value: ["leaf-12"]
+```
+
+NFD owns `.status.nodes`; Topograph owns only the desired CR specs and metadata.
+Group names should be stable and Kubernetes-safe, using a short hash when the
+raw topology value is too long or contains invalid name characters.
+
+## Engine Parameters
+
+Initial parameters:
+
+- `nodeSelector`: optional, same meaning as the `k8s` engine.
+- `namespace`: optional if the installed NFD CRDs are namespaced. The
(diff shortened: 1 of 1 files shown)


docs/design/nfd-engine-sdd.md
```diff
@@ -0,0 +1,221 @@
...
+NFD owns `.status.nodes`; Topograph owns only the desired CR specs and metadata.
+Group names should be stable and Kubernetes-safe, using a short hash when the
+raw topology value is too long or contains invalid name characters.
+
+## Engine Parameters
+
+Initial parameters:
+
+- `nodeSelector`: optional, same meaning as the `k8s` engine.
+- `namespace`: optional if the installed NFD CRDs are namespaced. The
```
ArangoGutierrez, line 110: NodeFeature and NodeFeatureGroup ship with a fixed `scope: Namespaced` in NFD's CRD manifest (only NodeFeatureRule is cluster-scoped), so CRD scope isn't an install-time variable to resolve via discovery or the REST mapper. Simpler to make `namespace` a required parameter with a sensible default (e.g. the NFD namespace).

ArangoGutierrez (commented):
Dmitry — strong writeup, and the Scheduling Caveat captures our conversation honestly. Ahead of our call I want to put the etcd sizing on firmer ground, and sketch a consumption model that stays inside the limits; it also points at what we should propose upstream to NFD.

## 1. The etcd constraint is a per-object cliff, not a storage budget

The number that bites is not the 2 GiB quota — it is etcd's default max request size of **1.5 MiB per write, i.e. per object**. Custom resources are stored as JSON, so each `status.nodes` entry costs roughly `len(nodeName) + 12` bytes (`{"name":"..."},`).

The dangerous object is the top of the tree: a core-switch group listing every node in the cluster.

| Node-name style | Name length | `status.nodes` at 10k nodes | % of 1.5 MiB cap |
|---|---:|---:|---:|
| HPC short (`node0421`) | 8 | ~200 KiB | 13% |
| Typical (`dgx-rack12-node03`) | 20 | ~320 KiB | 21% |
| EC2 internal DNS (`ip-10-0-123-45.us-west-2.compute.internal`) | 42 | ~540 KiB | 36% |
| GKE-style | 60 | ~720 KiB | 48% |
| Kubernetes max name (253) | 253 | ~2.6 MiB | **unwritable** |

Inverted: a single NodeFeatureGroup stops being writable at **~5,800 nodes** with max-length names, ~27,000 with EC2-style names, ~47,000 with 20-char names. So the 250-500 KiB estimate in this doc for one large group holds only for short names, and there is no headroom left at 100k.

Two notes on top of the raw cap:

- This applies to **any** whole-cluster NodeFeatureGroup, not just ours — a plain "all GPU nodes" group hits the same wall. That makes it an NFD problem worth fixing upstream, not a topograph workaround.
- The pain starts well before the cliff: every membership flap (autoscale, cordon, reboot) re-sends the **entire group object to every watcher**. A 540 KiB core group updating on every node event is operationally worse than the hard limit it has not hit yet.

## 2. What a scheduler actually needs at decision time

The scheduler already caches every Node in its informer snapshot — group **membership via the topology labels is already in its memory**, delivered incrementally (a node flap re-sends one ~5 KiB Node, not a half-MiB group). Kueue TAS reconstructs its whole topology tree from cached node labels today; that precedent is load-bearing.

The only data labels cannot carry: the **tree structure** (which leaf hangs off which spine) and the **group attributes** (the measured bandwidth/latency classes you proposed in the Slack thread). Both are O(switches), not O(nodes).

```mermaid
flowchart LR
    subgraph watch["watch time — rare, off the scheduling path"]
        CR["structure + attributes<br/>(~300 KiB total at 10k nodes)"] --> IDX["in-memory index<br/>groupID → tier, parent, bw class"]
    end
    subgraph sched["schedule time — per pod or gang"]
        N["NodeInfo.labels:<br/>leaf = leaf-12"] --> L1["index[leaf-12] →<br/>parent spine-2, bw class A"]
        L1 --> L2["index[spine-2] →<br/>parent core-1"]
    end
    IDX -.serves.-> L1
```

Same-leaf check = string compare; group distance = walk 1-3 parent pointers; "a leaf with >= K schedulable nodes" = a label histogram over the scheduler's own node cache. No API calls in the hot path, and nothing in etcd grows with node count.

## 3. Two encodings that stay inside the limits

Principle: etcd stores only what consumers cannot derive. Membership is derivable (labels); structure and attributes are not.

Using the doc's example fabric:

```mermaid
graph TD
    core1[core-1]
    core1 --- spine1[spine-1]
    core1 --- spine2[spine-2]
    spine1 --- leaf1["leaf-1<br/>n1..n4 (members)"]
    spine1 --- leaf2["leaf-2<br/>n5..n8 (members)"]
    spine2 --- leaf3["leaf-3<br/>n9..n12 (members)"]
    spine2 --- leaf4["leaf-4<br/>n13..n16 (members)"]
    style core1 fill:#f5f5f5,stroke:#999
    style spine1 fill:#f5f5f5,stroke:#999
    style spine2 fill:#f5f5f5,stroke:#999
```

Grey tiers = structure + attributes only, no member lists. Leaf tier = the only place membership materializes, bounded by switch radix (32-64 nodes), never by cluster size.

| Encoding | Largest object @10k (EC2 names) | @100k | Consumer path |
|---|---|---|---|
| Full membership at all tiers (current draft) | ~540 KiB (core NFG) | unwritable | read large lists |
| **Leaf-only membership + parent pointers** | 3-5 KiB (one leaf) | 3-5 KiB | walk parents |
| Labels + one structure/attributes object, no NFG membership | ~300 KiB (single CR) | shard per tier | map lookups |

The middle row keeps NFD in the loop and is expressible today with the metadata this doc already proposes: the `label-key`/`label-value` annotations plus one more, `topograph.nvidia.com/parent`, make the tree walkable without any set-containment inference. Spine/core groups define membership as the union of their children and never list nodes.

## 4. What this suggests upstream in NFD

Two API asks, both now justified by arithmetic rather than taste:

1. **Union rules** — already in this doc's Future Enhancements. Makes parentage first-class in spec and spares nfd-master re-evaluating per-node rules at every tier.
2. **A membership policy on NodeFeatureGroup** (`Full | Compressed | None`) — groups that exist for structure and attributes without materializing `status.nodes`; range compression as the middle setting where node naming is compressible. Given any whole-cluster group breaks at ~5.8k max-length names, NFD has this exposure independent of topograph.

We are planning to take the node-local half of this design to NFD as a KEP for a **network-topology-updater** worker (NIC/bond state, IMEX/NVLink clique identity, bandwidth/latency probes feeding the graph as NodeFeatures). The encoding above would be the etcd contract that keeps that proposal inside limits from day one.

## 5. Concrete suggestions for this doc

- **etcd section**: replace the 10-30 MiB aggregate framing with the per-object cliff (table above) and the watch-amplification note — the binding constraint is per-object, not total.
- **Proposed behavior**: consider leaf-only membership + parent metadata for upper tiers.
- **Scheduling caveat**: add the consumption model — labels answer "where is this node", the structure/attributes object answers "what is that place worth", and the plugin joins the two in memory.

Happy to walk through all of this on the call.

docs/design/nfd-engine-sdd.md
```diff
@@ -0,0 +1,221 @@
...
+  CRD scope.
+- `cleanup`: optional boolean, default `true`, deleting stale Topograph-managed
+  `NodeFeature` and `NodeFeatureGroup` objects no longer present in the latest
+  graph.
+
+## Implementation Notes
+
+- Add `pkg/engines/nfd` with the standard `NamedLoader`.
+- Register it in `pkg/registry/registry.go`.
+- Factor the current `k8s` label projection into a shared helper so both engines
```
ArangoGutierrez, line 121: Worth spelling out how this factoring handles the gpu.clique path: the k8s engine's projection reads `nvidia.com/gpu.clique` off live Node labels (`skipAcceleratorLabelWhenGPUCliqueExists`, #341), and slinky ended up reimplementing that logic separately (#342) rather than sharing it. The projection isn't derivable from `topology.Graph` alone, so the shared helper needs a node-labels input in its contract.

docs/design/nfd-engine-sdd.md
```diff
@@ -0,0 +1,221 @@
...
+- Compressed node membership in status. On large clusters, especially clusters
+  with more than 1000 nodes, listing every matched node in every group can make
+  `NodeFeatureGroup` status large and increase etcd storage pressure. NFD could
+  consider compact node-name ranges such as `node[100-199]` when node names are
+  compatible with range compression.
+
+## etcd Size Estimate
+
+For a 10,000-node cluster, the rough live custom-resource payload is likely in
+the 10-30 MiB range before etcd MVCC history, compaction effects, and filesystem
```
ArangoGutierrez, line 179: The 10-30 MiB aggregate isn't the binding constraint — etcd's ~1.5 MiB per-object request cap is. `status.nodes` costs ~`len(name)+12` bytes per entry as JSON: at 10k nodes a core-tier group is ~540 KiB with EC2-style names and ~2.6 MiB (unwritable) at the 253-char name limit; a single group stops being writable at ~5,800 max-length node names. Full arithmetic plus two in-limit encodings in the review summary.

PR #384, opened 2026-07-09
fix(chart): drop unused RBAC verbs from API server and node-data-broker ClusterRoles

ArangoGutierrez:
## Description

Removes provably-unused RBAC verbs from the Topograph API server and node-data-broker `ClusterRole`s, tightening them toward least privilege. No runtime behavior changes — every dropped verb has zero call sites in the codebase.

Dropped verbs:
- **API server (`topograph`)**: `pods: get`, `daemonsets: list`, `configmaps: list` (slinky engine rule)
- **node-data-broker**: `nodes: list`, `daemonsets: list` (IB rule), `pods: get` (IB rule)

Each removal was cross-referenced against actual `client-go` usage: the observer/broker/engine code paths use `List`/`Watch`/`Get`/`Update`/`Create` only where retained, and never exercise the dropped verbs (e.g. no `Pods().Get`, no `DaemonSets().List`, no `ConfigMaps().List` anywhere; the broker uses no informers and never lists nodes). The retained verbs are unchanged, so no supported provider/engine combination loses a grant it uses.

Helm-unittest coverage was extended with discriminating `contains`/`notContains` assertions in `rbac_test.yaml` and `subchart_rbac_test.yaml` (each mutation-verified to fail if a dropped verb reappears), and the render snapshot was refreshed.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](./CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] All commits are signed off per DCO (`git commit -s`).

PR #385, opened 2026-07-09
fix(chart): tighten node-observer ClusterRole (drop unused get, gate node access)

ArangoGutierrez:
## Description

Tightens the `node-observer` `ClusterRole` toward least privilege:

- **Drops the unused `get` verb.** The observer is entirely informer-driven (`SharedInformerFactory` → `List`+`Watch`); it never issues a direct `Get` on nodes or pods.
- **Gates cluster-wide `nodes` `list`/`watch` on `trigger.nodeSelector`.** The node informer is constructed only when a node trigger is configured (`pkg/node_observer/status_informer.go`), so a default install that only watches pods no longer receives cluster-wide node read/watch. `pods: [list, watch]` remain unconditional — the API-server and node-data-broker pod-readiness informers always run.

Also documents that `node-observer.rbac.create=false` suppresses the role entirely, and adds a Node Observer RBAC subsection to `docs/engines/k8s.md`.

Helm-unittest coverage asserts the full rules array (stronger than per-rule `contains`), verified both with and without `trigger.nodeSelector` set; the render snapshot was refreshed (observer rule only).

## Checklist


ArangoGutierrez:
@dmitsh companion to #384 — tightens the node-observer ClusterRole: drops the unused `get` (the observer is informer-driven, List+Watch only) and gates cluster-wide `nodes` list/watch on `trigger.nodeSelector` being set, since the node informer is only built then. `pods` list/watch stay (the readiness informers always run). Docs + CHANGELOG updated, snapshot refreshed, `make chart-test` green. Thanks for taking a look!

ArangoGutierrez:
Heads-up on the red `govulncheck` check here: it's **not** from this PR (which is chart-only). It's a freshly-published Go stdlib advisory — [GO-2026-5856](https://pkg.go.dev/vuln/GO-2026-5856), an ECH privacy leak in `crypto/tls` reachable from existing code — that turns CI red on `main` and every open PR. Fixed in Go 1.25.12; bumped in #390. Once #390 merges, a rebase here clears the check. Every other check (build, test, chart-test, DCO) is green.

dmitsh (approved):

ArangoGutierrez:
@dmitsh this one is approved and all presubmit checks are green after the rebase onto your 1.26.5 bump (build, test, chart-test, govulncheck all passing). Ready to merge whenever you like.

PR #390, opened 2026-07-10
fix: bump Go toolchain to 1.25.12 for GO-2026-5856 (crypto/tls)

ArangoGutierrez:
## Description

`govulncheck` in CI is failing on `main` and on every open PR because of a newly-published Go standard-library advisory, **[GO-2026-5856](https://pkg.go.dev/vuln/GO-2026-5856)** — an Encrypted Client Hello privacy leak in `crypto/tls`. It is reachable from `internal/httpreq`, `pkg/server/http_server.go`, and `pkg/engines/slurm`, so `govulncheck` exits non-zero. It is **fixed in the Go standard library at 1.25.12**.

This bumps the pinned toolchain `1.25.11 → 1.25.12` in the four places it appears (`go.mod` `go` directive, `Dockerfile` builder image, and the four `setup-go` pins in `.github/workflows/go.yml`), and refreshes the version reference in `AGENTS.md`, `.claude/CLAUDE.md`, and the existing `CHANGELOG.md` `[Unreleased]` entry.

No source changes; the same maintenance step as the previous `1.25.x` bumps.

### Verification

```
$ govulncheck ./...
No vulnerabilities found.
Your code is affected by 0 vulnerabilities.
```
`make qualify` passes (fmt, vet, lint, test).

This unblocks CI on `main` and on the open RBAC-hardening PRs (#384, #385).

## Checklist
- [x] I am familiar with the [Contributing Guidelines](./CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] All commits are signed off per DCO (`git commit -s`).

PR #393, opened 2026-07-10
feat(chart): add opt-in NetworkPolicy for the Topograph release

### LINKED ISSUE #383: Chart: optional, provider-aware NetworkPolicy template (opt-in)
## Summary

Add an optional, opt-in `NetworkPolicy` template to the Topograph Helm chart so the chart can provision the network access its own components need when deployed onto clusters that enforce a default-deny posture.

## Motivation

Some platforms employ a **global default-deny** security posture and expect each add-on chart installed afterward to bring the network policy its own workloads require, rather than the platform attempting to maintain a per-component carve-out centrally. Because the Topograph chart is generic (not platform-specific), a self-contained, opt-in policy keeps the chart portable while letting deny-all environments adopt it without editing central policy.

## Proposal

Add `networkPolicy.enabled` (default `false`) that renders a standard `networking.k8s.io/v1` `NetworkPolicy` (portable across CNIs that honor Kubernetes NetworkPolicy, e.g., Calico, Cilium). When enabled, it should cover both directions the components actually need:

**Ingress**
- To the Topograph API `Service` port (`49021` by default) from within the release namespace.
- From the Prometheus scraper namespace when `serviceMonitor.enabled: true`.

**Egress (provider-aware)**
- DNS (kube-dns) and the Kubernetes API server for all components.
- For the `netq` provider: egress to the configured NetQ endpoint (host/port from provider params) — the current doc snippet is ingress-only and does not cover this.
- `infiniband-k8s` / `dra` need no extra egress beyond DNS + API server.

Suggested shape (illustrative):

```yaml
networkPolicy:
  enabled: false
  # Namespace running Prometheus (used when serviceMonitor.enabled: true)
  prometheusNamespace: monitoring
  # Extra egress rules for providers that reach external endpoints (e.g. netq)
  extraEgress: []
```

## Notes / open questions

- **Policy-engine ordering:** on clusters where the deny-all is expressed as a Calico `GlobalNetworkPolicy` (rather than a namespaced Kubernetes `NetworkPolicy`), the interaction/precedence between a chart-shipped `NetworkPolicy` and a cluster `GlobalNetworkPolicy` should be validated — a chart-shipped allow may not override a GNP deny depending on tier/order. Worth documenting the expectation.
- Keep the template **opt-in and default-off** so it is inert on clusters that don't enforce NetworkPolicy.
- Scope is intentionally network-layer only. Pod-security admission (e.g., the `infiniband-k8s` broker running privileged with a `/sys/class` hostPath) is admission control and cannot be self-granted by a workload chart — that exemption necessarily remains an operator/platform action and is out of scope here.

resker:
## Description

Adds an opt-in `NetworkPolicy` (`networkPolicy.enabled`, default `false`) covering all three chart components, so the chart can provision the network access its own workloads need on clusters that enforce a default-deny posture. Closes #383.

The policy selects the whole release by instance label. When enabled:

- **Ingress** — denied except intra-release traffic (so the node-observer can reach the API server) and the Prometheus scrape namespace when `serviceMonitor.enabled`.
- **Egress** — left unconstrained until `extraEgress` is set, so enabling the policy cannot break egress on clusters without a default-deny. When `extraEgress` is set, an `Egress` policyType is added with DNS and intra-release allows plus the supplied rules.

Because Kubernetes NetworkPolicy egress cannot target hostnames, the kube-apiserver and the provider's topology-source endpoint are operator-supplied via `extraEgress` (required under default-deny). Rather than auto-derive these per provider (their endpoints are hostnames/URLs, not addressable in a portable policy), the chart stays generic and `docs/engines/k8s.md` includes a per-provider egress table. `extraIngress` appends custom ingress rules.

### Validation

- `make chart-test` green. helm-unittest covers off-by-default, the whole-release podSelector, the Prometheus ingress rule, and the egress shape.
- On a representative Kubernetes cluster (Calico CNI) with a namespaced default-deny NetworkPolicy: `helm test` succeeds under deny-all (release pods reach the API server), while a non-release pod is correctly blocked — confirming the intra-release allows are both sufficient and enforcing.

> Note: the red `govulncheck` check is the pre-existing repo-wide stdlib advisory (GO-2026-5856), not from this chart-only PR; it clears on a rebase once #390 lands. All other checks are green.

## Checklist


charts/topograph/values.yaml
```diff
@@ -212,6 +212,55 @@ serviceMonitor:
...
+  extraIngress: []
+  extraEgress: []
+  #  # Kubernetes API server (all providers under default-deny; adjust to your
+  #  # cluster -- `kubectl get endpoints kubernetes -n default`):
+  #  - to:
+  #      - ipBlock:
+  #          cidr: 10.96.0.1/32
+  #    ports:
+  #      - protocol: TCP
+  #        port: 6443
```
dmitsh: This example combines the Kubernetes Service IP with the usual control-plane endpoint port. In-cluster client-go uses KUBERNETES_SERVICE_HOST/KUBERNETES_SERVICE_PORT, normally the Service IP on 443; kubectl get endpoints returns an endpoint IP normally on 6443. Depending on where the CNI evaluates policy relative to DNAT, operators may need either Service-IP:443 or endpoint-IP:6443, but Service-IP:6443 matches neither path and can leave all three components unable to reach the API after Egress isolation is enabled. Please correct this example and explain the CNI-dependent alternative consistently with the docs example/table.

charts/topograph/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,73 @@
...
+    {{- if .Values.serviceMonitor.enabled }}
+    # Prometheus scrapes the API server /metrics on the service port.
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Values.serviceMonitor.namespace }}
+      ports:
+        - protocol: TCP
+          port: {{ .Values.service.port }}
+    {{- end }}
```
dmitsh, line 41: When ingress.enabled or gatewayAPI.enabled is used, requests normally originate from an ingress/gateway controller in another namespace. This policy permits only same-release pods and, conditionally, the ServiceMonitor namespace, so enabling networkPolicy makes those chart-provided routes unreachable unless the operator happens to add the correct extraIngress rule. Since the controller selector is cluster-specific, please at least document this required extraIngress configuration beside the Ingress/HTTPRoute guidance and add combination tests; otherwise two supported chart features silently conflict.

charts/topograph/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - from:
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/instance: {{ .Release.Name }}
+    {{- if .Values.serviceMonitor.enabled }}
+    # Prometheus scrapes the API server /metrics on the service port.
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ .Values.serviceMonitor.namespace }}
```
dmitsh: serviceMonitor.namespace controls where the ServiceMonitor object is created, not necessarily the namespace of the Prometheus pod that performs the scrape. A central Prometheus can watch ServiceMonitors in other namespaces, so this selector can still block metrics even though serviceMonitor.enabled is correctly configured. Please use a separately configurable scraper namespace/selector (or require extraIngress and document it) rather than treating the ServiceMonitor resource namespace as the traffic source.

charts/topograph/tests/networkpolicy_test.yaml
```diff
@@ -0,0 +1,97 @@
...
+          content: Ingress
+      - notContains:
+          path: spec.policyTypes
+          content: Egress
+      # Intra-release ingress so the node-observer can reach the API server.
+      - equal:
+          path: spec.ingress[0].from[0].podSelector.matchLabels["app.kubernetes.io/instance"]
+          value: chart-ci
+
+  - it: adds a Prometheus ingress rule when serviceMonitor is enabled
```
ArangoGutierrez: Two coverage gaps beyond dmitsh's notes: the extraIngress append branch has no test (only extraEgress is exercised, so deleting that template block leaves every case green), and this Prometheus rule asserts the source namespace but never the port — a regression on service.port would pass. An extraIngress case, ideally combined with serviceMonitor.enabled, would also cover the combination test dmitsh asked for.

charts/topograph/values.yaml
```diff
@@ -212,6 +212,55 @@ serviceMonitor:
...
+# cluster-specific and cannot be templated portably. `extraIngress` appends
+# custom ingress rules. Note: an empty rule object (`{}`) in extraIngress /
+# extraEgress matches ALL traffic in that direction -- always give explicit
+# `to`/`from`/`ports`.
+#
+# NetworkPolicy egress cannot target hostnames, so provider endpoints must be
+# given as ipBlock CIDRs. See docs/engines/k8s.md "Per-provider egress" for a
+# per-provider table (netq, CSP/IMDS caveat, dra/infiniband need none).
+networkPolicy:
+  enabled: false
```
ArangoGutierrez, line 266: Every other opt-in networking feature (ingress.enabled, gatewayAPI.enabled, nodeDataBroker.enabled) emits a NOTES.txt hint, but networkPolicy emits nothing. A conditional NOTE here would surface at install time exactly the silent footguns this introduces: egress needs extraEgress on default-deny clusters, ingress/gateway routes and external /v1/generate callers need extraIngress, and the policy is a no-op without an enforcing CNI.

ArangoGutierrez (commented):
Thorough, well-documented policy — podSelector coverage of all three components, the DNS empty-dict fallback, and the Prometheus port all hold up under helm-unittest. dmitsh's open comments already cover the apiserver-egress example, the ingress/gateway conflict, and the Prometheus-namespace source; two additive points those don't touch:

1. Two coverage gaps beyond dmitsh's notes: the extraIngress append branch has no test (only extraEgress is exercised, so deleting that template block leaves every case green), and this Prometheus rule asserts the source namespace but never the port — a regression on service.port would pass. An extraIngress case, ideally combined with serviceMonitor.enabled, would also cover the combination test dmitsh asked for. (charts/topograph/tests/networkpolicy_test.yaml:34)
2. Every other opt-in networking feature (ingress.enabled, gatewayAPI.enabled, nodeDataBroker.enabled) emits a NOTES.txt hint, but networkPolicy emits nothing. A conditional NOTE here would surface at install time exactly the silent footguns this introduces: egress needs extraEgress on default-deny clusters, ingress/gateway routes and external /v1/generate callers need extraIngress, and the policy is a no-op without an enforcing CNI. (charts/topograph/values.yaml:236)

resker:
Re: the Greptile/CodeRabbit schema feedback — both bots are flagging that `extraIngress`/`extraEgress` items can omit `from`/`to`, which Kubernetes treats as "all peers."

That's true, but it's intentional. A rule with only `ports` (no `from`) means "allow any source on these ports," and a rule with only `from` (no `ports`) means "allow this source on all ports" — both are valid and useful NetworkPolicy patterns. Banning them from the Helm schema would break legitimate operator configs.

What the schema *does* catch is the actually dangerous case: a bare `{}` or `{from: []}` that matches everything unintentionally. The `minProperties: 1` + `minItems: 1` constraints close those.

Peer-semantic validation ("this rule is broader than you may have intended") is really a job for an admission webhook or Kyverno policy, not Helm schema. The docs already call out the footgun explicitly: *"an empty rule object (`{}`) matches all traffic in that direction."*

ArangoGutierrez (approved):
Re-reviewed at 5e2d7282. Both points from my July 21 review are addressed, and I mutation-tested them rather than trusting the green run: deleting the `extraIngress` block turns the new case red (`count: 2` -> `1`), and swapping `.Values.service.port` for a literal fails the metrics port assertion. Real guards.

I also checked the things a NetworkPolicy usually gets wrong. The port chain is correct (`containerPort` -> `targetPort: http` -> `service.port`), which matters here because the policy matches the pod port, not the Service port. Rendering the full chart confirms all three components carry `app.kubernetes.io/instance`, so the podSelector genuinely covers the release, and the `helm test` hook pods carry it too, so `helm test` still passes with the policy on. Full suite is green at 147 tests / 153 snapshots.

The shape is right: off by default, and gating the `Egress` policyType on `extraEgress` being set means enabling this cannot break egress on a cluster that is not already default-deny. The per-provider egress table and the hostname/`ipBlock` caveat are the right level of detail, and calling out the IMDS credential-exposure risk is a good catch.

On the recurring bot comment about `extraIngress`/`extraEgress` accepting under-scoped rules: I looked at tightening the schema and decided against asking for it. Requiring `from`/`to` rejects `[{ports: [...]}]` but still admits `from: [{namespaceSelector: {}}]`, which is every namespace in the cluster, just as open. JSON Schema cannot express "this rule is meaningfully scoped", so it is a lint on operator misuse rather than a boundary, and the values.yaml and docs already warn about it explicitly. Not worth holding the PR.

LGTM. Two follow-ups for a later PR, neither blocking:

1. The combination tests dmitsh asked for never landed. No case sets `ingress.enabled` or `gatewayAPI.enabled` alongside `networkPolicy.enabled`. The docs and the NOTES warning cover the footgun for humans, but nothing pins it in CI, so a future change to either template can silently reintroduce the conflict.
2. The new NOTES entry and the ServiceAccount entry are both numbered `2.`, and NOTES does not mention the policy is inert without an enforcing CNI (docs/engines/k8s.md does). That is the failure mode people hit silently.

PR #398, opened 2026-07-13
refactor(helm): fold node components into main chart

dmitsh:
Remove the node-observer and node-data-broker subcharts, flatten shared
values, and reuse root image, naming, pull-secret, and verbosity settings.


charts/topograph/tests/subchart_rbac_test.yaml
```diff
@@ -1,13 +1,13 @@
-suite: subchart rbac
```
giuliocalzo, line 1: @dmitsh we can even have test for every component , instead then having a generic

`tests/component_rbac_test.yaml` we can have

`tests/node-observer_rbac_test.yaml` and `tests/node-data-broker_rbac_test.yaml

giuliocalzo (approved):
LGTM

PR #402, opened 2026-07-14
feat(engine/nfd): publish topology as NFD custom resources

dmitsh:
Add an NFD engine that translates the canonical topology graph into
NodeFeature and NodeFeatureGroup resources.

Publish accelerator, leaf, spine, and core topology attributes, including the
system node name required for NFD to populate group membership. Support node
selection, namespaced NFD installations, stale-resource cleanup, and existing
GPU clique labels.

### DIFF
--- a/charts/topograph/templates/rbac.yaml
+++ b/charts/topograph/templates/rbac.yaml
@@ -49,6 +49,11 @@ rules:
   resources: [configmaps]
   verbs: [create,get,update]
 {{- end }}
+{{- if eq .Values.engine.name "nfd" }}
+- apiGroups: [nfd.k8s-sigs.io]
+  resources: [nodefeatures,nodefeaturegroups]
+  verbs: [create,get,list,patch,update,delete]
+{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/charts/topograph/templates/rbac.yaml
+++ b/charts/topograph/templates/rbac.yaml
@@ -49,11 +49,6 @@ rules:
   resources: [configmaps]
   verbs: [create,get,update]
 {{- end }}
-{{- if eq .Values.engine.name "nfd" }}
-- apiGroups: [nfd.k8s-sigs.io]
-  resources: [nodefeatures,nodefeaturegroups]
-  verbs: [create,get,list,patch,update,delete]
-{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -68,4 +63,31 @@ roleRef:
   kind: ClusterRole
   name: {{ include "topograph.rbacName" . }}
   apiGroup: ""
+{{- if eq .Values.engine.name "nfd" }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "topograph.rbacName" . }}
+  namespace: {{ .Values.nfdNamespace | quote }}
+rules:
+- apiGroups: [nfd.k8s-sigs.io]
+  resources: [nodefeatures,nodefeaturegroups]
+  verbs: [create,list,patch,delete]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "topograph.rbacName" . }}
+  namespace: {{ .Values.nfdNamespace | quote }}
+subjects:
+- kind: ServiceAccount
+  name: {{ include "topograph.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  apiGroup: ""
+roleRef:
+  kind: Role
+  name: {{ include "topograph.rbacName" . }}
+  apiGroup: ""
+{{- end }}
 {{- end }}
--- a/charts/topograph/tests/rbac_test.yaml
+++ b/charts/topograph/tests/rbac_test.yaml
@@ -71,7 +71,7 @@ tests:
             resources: [daemonsets]
             verbs: [get]
 
-  - it: does not grant pods/exec or configmaps with the default provider and engine
+  - it: does not grant provider- or engine-specific access by default
     documentIndex: 0
     asserts:
       - notContains:
@@ -85,7 +85,13 @@ tests:
           content:
             apiGroups: [""]
             resources: [configmaps]
-            verbs: [create, get, list, update]
+            verbs: [create, get, update]
+      - notContains:
+          path: rules
+          content:
+            apiGroups: [nfd.k8s-sigs.io]
+            resources: [nodefeatures, nodefeaturegroups]
+            verbs: [create, list, patch, delete]
       - notContains:
           path: rules
           content:
@@ -135,14 +141,60 @@ tests:
     set:
       engine:
         name: nfd
-    documentIndex: 0
     asserts:
+      - hasDocuments:
+          count: 4
+      - notContains:
+          path: rules
+          content:
+            apiGroups: [nfd.k8s-sigs.io]
+            resources: [nodefeatures, nodefeaturegroups]
+            verbs: [create, list, patch, delete]
+        documentIndex: 0
+      - isKind:
+          of: Role
+        documentIndex: 2
+      - equal:
+          path: metadata.namespace
+          value: node-feature-discovery
+        documentIndex: 2
       - contains:
           path: rules
           content:
             apiGroups: [nfd.k8s-sigs.io]
             resources: [nodefeatures, nodefeaturegroups]
-            verbs: [create, get, list, patch, update, delete]
+            verbs: [create, list, patch, delete]
+        documentIndex: 2
+      - isKind:
+          of: RoleBinding
+        documentIndex: 3
+      - equal:
+          path: metadata.namespace
+          value: node-feature-discovery
+        documentIndex: 3
+      - equal:
+          path: roleRef.kind
+          value: Role
+        documentIndex: 3
+      - equal:
+          path: subjects[0].namespace
+          value: topograph
+        documentIndex: 3
+
+  - it: scopes NFD CR access to the configured namespace
+    set:
+      engine:
+        name: nfd
+      nfdNamespace: custom-nfd
+    asserts:
+      - equal:
+          path: metadata.namespace
+          value: custom-nfd
+        documentIndex: 2
+      - equal:
+          path: metadata.namespace
+          value: custom-nfd
+        documentIndex: 3
 
   - it: does not grant pods/exec for slinky with a default flat cluster topology
     set:
--- a/charts/topograph/tests/rbac_test.yaml
+++ b/charts/topograph/tests/rbac_test.yaml
@@ -131,6 +131,19 @@ tests:
             resources: [configmaps]
             verbs: [create, get, list, update]
 
+  - it: grants NFD CR access for the nfd engine
+    set:
+      engine:
+        name: nfd
+    documentIndex: 0
+    asserts:
+      - contains:
+          path: rules
+          content:
+            apiGroups: [nfd.k8s-sigs.io]
+            resources: [nodefeatures, nodefeaturegroups]
+            verbs: [create, get, list, patch, update, delete]
+
   - it: does not grant pods/exec for slinky with a default flat cluster topology
     set:
       engine:
--- a/charts/topograph/tests/rbac_test.yaml
+++ b/charts/topograph/tests/rbac_test.yaml
@@ -137,6 +137,20 @@ tests:
             resources: [configmaps]
             verbs: [create, get, list, update]
 
+  - it: does not render NFD namespace RBAC for an explicit non-NFD engine
+    set:
+      engine:
+        name: slinky
+    asserts:
+      - hasDocuments:
+          count: 2
+      - isKind:
+          of: ClusterRole
+        documentIndex: 0
+      - isKind:
+          of: ClusterRoleBinding
+        documentIndex: 1
+
   - it: grants NFD CR access for the nfd engine
     set:
       engine:
--- a/pkg/engines/nfd/engine.go
+++ b/pkg/engines/nfd/engine.go
@@ -1,17 +1,6 @@
 /*
- * Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * Copyright 2026 NVIDIA CORPORATION
+ * SPDX-License-Identifier: Apache-2.0
  */
 
 package nfd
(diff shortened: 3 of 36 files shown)


pkg/engines/nfd/engine.go
```diff
@@ -0,0 +1,156 @@
...
+	params        *Params
+	cachedNodes   *corev1.NodeList
+}
+
+type Params struct {
+	// NodeSelector (optional) specifies nodes participating in the topology.
+	NodeSelector map[string]string `mapstructure:"nodeSelector"`
+	// Namespace (optional) is used only for NFD installs with namespaced CRDs.
+	// NFD NodeFeature and NodeFeatureGroup are normally cluster-scoped.
+	Namespace string `mapstructure:"namespace"`
```
ArangoGutierrez: NodeFeature and NodeFeatureGroup are namespaced in NFD v0.18.3 (only NodeFeatureRule is cluster-scoped), so this comment is inverted, and with the default empty namespace the dynamic client issues cluster-scoped calls that fail against a real API server; the fake client in the tests does not enforce scope, so the suite stays green. Consider making namespace required or defaulted, as suggested in the #382 design review, or deriving the scope from discovery/the REST mapper as the SDD describes.

pkg/engines/nfd/objects.go
```diff
@@ -0,0 +1,430 @@
...
+			"labels":      obj.GetLabels(),
+			"annotations": obj.GetAnnotations(),
+		},
+		"spec": obj.Object["spec"],
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = resource.Patch(ctx, obj.GetName(), types.MergePatchType, data, metav1.PatchOptions{})
```
ArangoGutierrez: JSON merge patch only overwrites keys present in the patch document, so when a node loses a topology attribute between runs (accelerator, leaf, spine, or core), the stale key persists in the live NodeFeature elements map indefinitely; cleanup only removes whole objects. Server-side apply, a JSON patch that replaces the spec, or emitting the full key set with explicit removals would make reconciliation converge.

charts/topograph/templates/rbac.yaml
```diff
@@ -49,6 +49,11 @@ rules:
   resources: [configmaps]
   verbs: [create,get,update]
 {{- end }}
+{{- if eq .Values.engine.name "nfd" }}
+- apiGroups: [nfd.k8s-sigs.io]
+  resources: [nodefeatures,nodefeaturegroups]
+  verbs: [create,get,list,patch,update,delete]
```
ArangoGutierrez: The nfd package only calls create, patch, list, and delete on these resources; get and update are unused. Dropping them keeps the rule consistent with the least-privilege cleanup from #384.

ArangoGutierrez (commented):
Found 3 issues.

1. NodeFeature and NodeFeatureGroup are namespaced in NFD v0.18.3 (only NodeFeatureRule is cluster-scoped), so this comment is inverted, and with the default empty namespace the dynamic client issues cluster-scoped calls that fail against a real API server; the fake client in the tests does not enforce scope, so the suite stays green. Consider making namespace required or defaulted, as suggested in the #382 design review, or deriving the scope from discovery/the REST mapper as the SDD describes. (pkg/engines/nfd/engine.go:54)
2. JSON merge patch only overwrites keys present in the patch document, so when a node loses a topology attribute between runs (accelerator, leaf, spine, or core), the stale key persists in the live NodeFeature elements map indefinitely; cleanup only removes whole objects. Server-side apply, a JSON patch that replaces the spec, or emitting the full key set with explicit removals would make reconciliation converge. (pkg/engines/nfd/objects.go:310)
3. The nfd package only calls create, patch, list, and delete on these resources; get and update are unused. Dropping them keeps the rule consistent with the least-privilege cleanup from #384. (charts/topograph/templates/rbac.yaml:55)

ArangoGutierrez:
Ran this branch end-to-end on kind v0.32 with NFD v0.18.3 (`featureGates.NodeFeatureGroupAPI=true`), `provider: test`, `engine: nfd`. Results, substantiating two of the review comments and confirming the design works otherwise:

**Namespace default (review comment on `engine.go`):** reproduced. `nodefeatures`/`nodefeaturegroups` are `scope: Namespaced` in NFD v0.18.3, so with the default empty namespace every apply fails with `the server could not find the requested resource`. Because the engine wraps this as 502 and `httpreq.ShouldRetry` treats 502 as retryable, each generate request runs the full 2s/4s/8s/16s backoff cycle before surfacing a deterministic failure. With `engine.params.namespace` set to the NFD namespace everything works.

**Merge-patch staleness (review comment on `objects.go`):** reproduced. Generated from a model where nodes 1201/1202 lose their accelerator domain between runs: the `accelerator-nvl2` NodeFeatureGroup is correctly deleted by cleanup, but the NodeFeature for 1201 still carries `accelerator: nvl2` afterwards — the merge patch never removes absent keys, leaving an attribute that references a group that no longer exists.

**The good news:** with the namespace set, the full pipeline works. 8 NodeFeatures + 11 NodeFeatureGroups from the `medium.yaml` model matched expectations exactly, and NFD master populated `status.nodes` correctly for all 11 groups on bare Node objects with no NFD worker running — the `system.name/nodename` mechanism does what the docs claim for KWOK-style nodes. One boundary worth documenting: NFD silently skips NodeFeatures whose node does not exist as a Kubernetes Node object (`node not found, skip update`), so the KWOK claim holds only because KWOK creates Node objects.

Repro recipe: kind cluster, NFD 0.18.3 chart with the group API gate, this branch's image, `provider: test` with the embedded models, plus bare `kind: Node` objects for the model hostnames. This would make a cheap CI e2e for the engine — happy to share the exact commands.

pkg/registry/registry.go
```diff
@@ -62,6 +63,7 @@ var Providers = providers.NewRegistry(
 
 var Engines = engines.NewRegistry(
 	k8s.NamedLoader,
+	nfd.NamedLoader,
```
ArangoGutierrez, line 66: This PR adds a fourth engine, but AGENTS.md and .claude/CLAUDE.md still describe three: the repository map says engines are k8s, slinky, slurm, and the Adding-a-new-engine section says three exist. Per the Documentation Impact table both files need updating in the same PR.

charts/topograph/templates/rbac.yaml
```diff
@@ -49,6 +49,11 @@ rules:
   resources: [configmaps]
   verbs: [create,get,update]
 {{- end }}
+{{- if eq .Values.engine.name "nfd" }}
```
ArangoGutierrez, line 66: NodeFeature and NodeFeatureGroup are namespaced and the engine only ever writes to the single configured engine.params.namespace, so a namespaced Role/RoleBinding scoped to that namespace would match the actual access pattern better than extending the shared ClusterRole cluster-wide - especially since this adds the first delete verb to it.

pkg/engines/nfd/engine.go
```diff
@@ -0,0 +1,161 @@
...
+	if eng.params.Cleanup {
+		if err := eng.cleanupObjects(ctx, nodeFeatures, nodeFeatureGroups); err != nil {
+			return nil, httperr.NewError(http.StatusBadGateway, err.Error())
+		}
+	}
+
+	return fmt.Appendf(nil, "OK nodeFeatures=%d nodeFeatureGroups=%d\n", len(nodeFeatures), len(nodeFeatureGroups)), nil
+}
+
+func gpuCliqueValues(nodes *corev1.NodeList) map[string]string {
```
ArangoGutierrez, line 150: This is the third independent implementation of projecting nvidia.com/gpu.clique from live node labels, after the k8s engine's skipAcceleratorLabelWhenGPUCliqueExists and slinky's withGPUCliqueDomains. The SDD review on #382 asked for a shared helper with node labels in its contract; internal/k8s/utils.go looks like the natural home.

charts/topograph/tests/rbac_test.yaml
```diff
@@ -131,6 +131,19 @@ tests:
             resources: [configmaps]
             verbs: [create, get, list, update]
 
+  - it: grants NFD CR access for the nfd engine
```
ArangoGutierrez, line 154: This asserts the rule is present when engine.name is nfd, but nothing asserts its absence otherwise - deleting the engine-name guard in rbac.yaml would leave the whole suite green. A notContains check in the default-engine negative test above would make the guard load-bearing.

ArangoGutierrez (commented):
Round-1 comments are addressed and check out at this head: the namespace scoping is defaulted and applied on every dynamic-client path, the merge patch is now a JSON Patch that drops stale topology keys (with a regression test), and the RBAC verbs match the client calls exactly. Found 4 issues, mostly around the edges.

1. This PR adds a fourth engine, but AGENTS.md and .claude/CLAUDE.md still describe three: the repository map says engines are k8s, slinky, slurm, and the Adding-a-new-engine section says three exist. Per the Documentation Impact table both files need updating in the same PR. (pkg/registry/registry.go:66)
2. NodeFeature and NodeFeatureGroup are namespaced and the engine only ever writes to the single configured engine.params.namespace, so a namespaced Role/RoleBinding scoped to that namespace would match the actual access pattern better than extending the shared ClusterRole cluster-wide - especially since this adds the first delete verb to it. (charts/topograph/templates/rbac.yaml:52)
3. This is the third independent implementation of projecting nvidia.com/gpu.clique from live node labels, after the k8s engine's skipAcceleratorLabelWhenGPUCliqueExists and slinky's withGPUCliqueDomains. The SDD review on #382 asked for a shared helper with node labels in its contract; internal/k8s/utils.go looks like the natural home. (pkg/engines/nfd/engine.go:148)
4. This asserts the rule is present when engine.name is nfd, but nothing asserts its absence otherwise - deleting the engine-name guard in rbac.yaml would leave the whole suite green. A notContains check in the default-engine negative test above would make the guard load-bearing. (charts/topograph/tests/rbac_test.yaml:134)

pkg/engines/nfd/engine.go
```diff
@@ -0,0 +1,170 @@
...
+	nodeFeatures, nodeFeatureGroups, err := buildNFDObjects(nodeLabels, gpuCliqueValues(nodes))
+	if err != nil {
+		return nil, httperr.NewError(http.StatusBadRequest, err.Error())
+	}
+
+	if err := eng.applyObjects(ctx, nodeFeatures, nodeFeatureGroups); err != nil {
+		return nil, httperr.NewError(http.StatusBadGateway, err.Error())
+	}
+
+	if eng.params.Cleanup {
```
ArangoGutierrez, line 141: GenerateOutput runs the cleanup pass whenever cleanup is enabled (the default), even when the graph is nil or yields zero objects: the desired set is empty, so every previously published NodeFeature and NodeFeatureGroup in the namespace gets deleted. A transient provider hiccup or an over-narrow node selector would wipe live topology data schedulers depend on. Treating an empty build result as an error (as slinky does for zero selected nodes) or skipping cleanup when nothing was generated would bound the blast radius; TestGenerateOutputCleansStaleObjects exercises exactly this path with a nil graph.

charts/topograph/tests/rbac_test.yaml
```diff
@@ -85,7 +85,13 @@ tests:
           content:
             apiGroups: [""]
             resources: [configmaps]
-            verbs: [create, get, list, update]
+            verbs: [create, get, update]
+      - notContains:
```
ArangoGutierrez, line 89: These notContains assertions check documentIndex 0, the shared ClusterRole, but the nfd rules render as a separate Role/RoleBinding at documentIndex 2/3, so deleting the engine-name guard in rbac.yaml leaves them green: the assertion cannot fail for the mutation it was added to catch (the same pattern repeats at line 147). Only the pre-existing default-engine hasDocuments count would trip, and only on full guard deletion. Asserting document count or Role absence for an explicit non-nfd engine would make the guard load-bearing.

pkg/engines/nfd/engine_test.go
```diff
@@ -0,0 +1,349 @@
...
+		staleGroup,
+	)
+	eng := &NfdEngine{
+		client:        k8sfake.NewSimpleClientset(),
+		dynamicClient: dynamicClient,
+		params:        &Params{Cleanup: true},
+		namespace:     testNFDNamespace,
+	}
+
+	out, httpErr := eng.GenerateOutput(context.Background(), nil, nil)
```
ArangoGutierrez: The cleanup test only covers the all-stale case: a nil graph empties the desired set and everything is deleted. A fixture with one still-desired and one stale object in the same run would exercise cleanupResource's retain branch, which is the realistic topology-changed case this cleanup exists for.

ArangoGutierrez (commented):
Round-2 follow-ups largely check out at this head: AGENTS.md and .claude/CLAUDE.md now describe four engines, and the NFD RBAC moved to a namespaced Role/RoleBinding whose verbs match the client calls exactly. Found 3 issues; the shared gpu.clique helper from round 2 remains open in its original thread.

1. GenerateOutput runs the cleanup pass whenever cleanup is enabled (the default), even when the graph is nil or yields zero objects: the desired set is empty, so every previously published NodeFeature and NodeFeatureGroup in the namespace gets deleted. A transient provider hiccup or an over-narrow node selector would wipe live topology data schedulers depend on. Treating an empty build result as an error (as slinky does for zero selected nodes) or skipping cleanup when nothing was generated would bound the blast radius; TestGenerateOutputCleansStaleObjects exercises exactly this path with a nil graph. (pkg/engines/nfd/engine.go:148)
2. These notContains assertions check documentIndex 0, the shared ClusterRole, but the nfd rules render as a separate Role/RoleBinding at documentIndex 2/3, so deleting the engine-name guard in rbac.yaml leaves them green: the assertion cannot fail for the mutation it was added to catch (the same pattern repeats at line 147). Only the pre-existing default-engine hasDocuments count would trip, and only on full guard deletion. Asserting document count or Role absence for an explicit non-nfd engine would make the guard load-bearing. (charts/topograph/tests/rbac_test.yaml:89)
3. The cleanup test only covers the all-stale case: a nil graph empties the desired set and everything is deleted. A fixture with one still-desired and one stale object in the same run would exercise cleanupResource's retain branch, which is the realistic topology-changed case this cleanup exists for. (pkg/engines/nfd/engine_test.go:196)

ArangoGutierrez (commented):
All three round-3 fixes check out at this head: the empty-set cleanup guard, the load-bearing non-nfd RBAC negative test, and the retain-plus-delete cleanup fixture. One expected string drifted between the new test and the implementation and is the only thing keeping the test job red; good to approve once CI is green.

1. This expected substring drifted from the implementation: GenerateOutput returns "generated no NFD topology objects; keeping the existing topology", so this assertion is the single failure in the test job. The implementation wording matches the docs ("preserves the existing topology"), so updating the expected string here looks like the right side to change. (pkg/engines/nfd/engine_test.go:262)

pkg/engines/nfd/engine_test.go
```diff
@@ -0,0 +1,426 @@
...
+	require.Nil(t, out)
+	require.Equal(t, http.StatusBadGateway, httpErr.Code())
+	require.ErrorContains(t, httpErr, "generated no NFD topology objects; refusing to clean up")
```
ArangoGutierrez: This expected substring drifted from the implementation: GenerateOutput returns "generated no NFD topology objects; keeping the existing topology", so this assertion is the single failure in the test job. The implementation wording matches the docs ("preserves the existing topology"), so updating the expected string here looks like the right side to change.

ArangoGutierrez (approved):
All follow-ups resolved: the empty-set cleanup guard has matching test coverage (retain+delete and cleanup-disabled paths included), the non-nfd RBAC negative test is now load-bearing, and CI is fully green including both codecov gates. The namespaced Role matches the engine's write pattern exactly and the NodeFeatureGroup namespace constraint is documented. LGTM.

ArangoGutierrez:
@dmitsh approved — ready for merge. Thanks for the quick turnarounds across the review rounds.

PR #403, opened 2026-07-15
fix(aws): use default credential chain

olegleyz:
## Description

The AWS provider currently calls the EC2 instance-role credential provider whenever Topograph-specific credentials are absent. That bypasses the AWS SDK default credential chain, so EKS Pod Identity and IRSA credentials injected into the Topograph pod are ignored and the EC2 node role is used instead.

This change:

- keeps request/config-file credentials as the highest-priority explicit override;
- leaves the credentials provider unset otherwise, allowing the AWS SDK to resolve environment, shared config, web identity, container/Pod Identity, and EC2 IMDS credentials;
- preserves the SDK credentials cache so temporary credentials refresh automatically;
- documents the default chain, EKS Pod Identity setup, and the least-privilege `ec2:DescribeInstanceTopology` permission.

Tests cover invalid and valid explicit credentials, explicit-over-environment precedence, environment credentials, EKS container credentials with an authorization token file, and refresh after expiration.

## Validation

- `make qualify`
- AWS provider race tests repeated 10 times

### AWS integration test

Validated commit `4473a9e` on the `nkx-slinky-gb300-dev-01` EKS cluster:

1. Deployed an isolated Topograph API using the `topograph` ServiceAccount and its EKS Pod Identity association with the dedicated `aehiqzudem-topograph` IAM role.
2. Set `AWS_EC2_METADATA_DISABLED=true` to prevent fallback to the EC2 node role.
3. Triggered a full topology generation request.

The request used the AWS SDK default credential chain, extracted topology for 35 instances, updated the test ConfigMap, and completed with HTTP 200. No authorization errors occurred.

## Checklist


pkg/providers/aws/provider.go
```diff
@@ -111,39 +98,32 @@ func Loader(ctx context.Context, cfg providers.Config) (providers.Provider, *htt
...
+	if credsProvider != nil {
+		opts = append(opts, config.WithCredentialsProvider(credsProvider))
 	}
 
-	return &Credentials{
-		AccessKeyId:     accessKeyID,
-		SecretAccessKey: secretAccessKey,
-		Token:           sessionToken,
-	}, nil
+	return config.LoadDefaultConfig(ctx, opts...)
```
ArangoGutierrez, line 128: One precedence change worth confirming is intentional: the old resolver only consulted request creds, then AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY, then the EC2 instance role — it never read ~/.aws/config, ~/.aws/credentials, or AWS_PROFILE. The full default chain places shared-config/profile (and IRSA) ahead of the instance role, so a host with stray ~/.aws state or an exported AWS_PROFILE that previously fell through to instance-role creds will now silently authenticate as that profile identity. The docs cover the env / Pod Identity case; worth extending the caveat to shared-config/profile too.

pkg/providers/aws/provider.go
```diff
@@ -111,39 +98,32 @@ func Loader(ctx context.Context, cfg providers.Config) (providers.Provider, *htt
 	return New(clientFactory, trimTiers), nil
 }
 
-func getCredentials(ctx context.Context, creds map[string]any) (*Credentials, *httperr.Error) {
-	var accessKeyID, secretAccessKey, sessionToken string
+func getCredentialsProvider(creds map[string]any) (aws.CredentialsProvider, *httperr.Error) {
+	if len(creds) == 0 {
+		klog.Infof("Using AWS SDK default credential chain")
+		return nil, nil
```
ArangoGutierrez, line 106: With no explicit creds, credential-resolution failure is now deferred to the API call: the old ec2rolecreds path returned 401 at load time when instance-role creds couldn't be retrieved, whereas an unresolvable default chain now surfaces as the 502 from DescribeInstanceTopology. #177 deliberately curated these status codes, so worth deciding whether a no-credentials failure should still present as 401 rather than 502.

ArangoGutierrez (commented):
Solid refactor — moving to the SDK default credential chain is the right call, and the new tests genuinely exercise precedence and auto-refresh rather than asserting SDK tautologies. Two behavior changes worth a quick confirm below.

1. One precedence change worth confirming is intentional: the old resolver only consulted request creds, then AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY, then the EC2 instance role — it never read ~/.aws/config, ~/.aws/credentials, or AWS_PROFILE. The full default chain places shared-config/profile (and IRSA) ahead of the instance role, so a host with stray ~/.aws state or an exported AWS_PROFILE that previously fell through to instance-role creds will now silently authenticate as that profile identity. The docs cover the env / Pod Identity case; worth extending the caveat to shared-config/profile too. (pkg/providers/aws/provider.go:126)
2. With no explicit creds, credential-resolution failure is now deferred to the API call: the old ec2rolecreds path returned 401 at load time when instance-role creds couldn't be retrieved, whereas an unresolvable default chain now surfaces as the 502 from DescribeInstanceTopology. #177 deliberately curated these status codes, so worth deciding whether a no-credentials failure should still present as 401 rather than 502. (pkg/providers/aws/provider.go:104)

pkg/providers/aws/provider.go
olegleyz: Yes, the expanded precedence is intentional because the goal is to use the standard SDK chain. I’ll make the warning explicit for AWS_PROFILE and shared config/credentials files.

pkg/providers/aws/provider.go
olegleyz: Good catch. The status-code change was not intentional. I’ll preserve the previous 401 for credential-resolution failures while keeping EC2 API authorization errors as 502.

ArangoGutierrez (commented):
Second-round review — both items from the previous round are addressed:

1. The credential-precedence caveat now documents that environment and shared-profile sources (AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY, AWS_PROFILE, ~/.aws/config, ~/.aws/credentials) can take priority over EKS Pod Identity or an EC2 instance role.
2. A missing/unresolvable credential now surfaces as 401 again, via the pre-flight Retrieve check before the EC2 call, with tests covering the 401 and 502 paths plus a real credential-refresh test against a fake container-credentials endpoint.

No new issues found — checked for bugs, Go conventions, test quality, security, and CLAUDE.md compliance.

dmitsh (approved):

PR #405, opened 2026-07-16
fix(provider/dra): match nodes by annotated instance ID

dmitsh:


pkg/providers/dra/provider.go
```diff
@@ -109,8 +109,13 @@ func (p *Provider) GenerateTopologyConfig(ctx context.Context, _ *int, instances
 		}
 
 		i2n := instances[indx].Instances
-		if host, ok := i2n[node.Name]; ok {
-			domainMap.AddHost(clusterID, node.Name, host)
+		instanceID, ok := node.Annotations[topology.KeyNodeInstance]
+		if !ok || instanceID == "" {
```
ArangoGutierrez, line 114: A clique-labeled node missing this annotation is now skipped silently. The sibling join sites warn on the same skip — internal/k8s/utils.go and engines/slinky/engine.go both do `klog.Warningf("missing %q annotation in node %s", ...)`, and #379 added exactly this diagnostic to the structurally identical clique+annotation loop in Slinky. A klog.Warningf before the continue would restore that signal; otherwise a not-yet-annotated node disappears with no trace, since the BadGateway error only fires when every node is skipped.

ArangoGutierrez (commented):
Correct fix — DRA was the last provider still joining on node.Name, and the new test genuinely discriminates it (node name differs from the instance ID). One optional diagnostic suggestion below.

1. A clique-labeled node missing this annotation is now skipped silently. The sibling join sites warn on the same skip — internal/k8s/utils.go and engines/slinky/engine.go both do `klog.Warningf("missing %q annotation in node %s", ...)`, and #379 added exactly this diagnostic to the structurally identical clique+annotation loop in Slinky. A klog.Warningf before the continue would restore that signal; otherwise a not-yet-annotated node disappears with no trace, since the BadGateway error only fires when every node is skipped. (pkg/providers/dra/provider.go:113)

PR #392, opened 2026-07-10
feat(chart)!: harden default pod security context to restricted PSS

ArangoGutierrez:
## Description

Populate the previously-empty `podSecurityContext` / `securityContext` defaults across all three chart workloads (API server, node-observer, node-data-broker) so the chart satisfies the Kubernetes **restricted** Pod Security Standard out of the box: `runAsNonRoot` (UID/GID 65532), `seccompProfile: RuntimeDefault`, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, and all capabilities dropped.

Closes #387.

**Authorship:** the hardening commit is authored by **@resker** (Rob Esker), who built and designed it as part of the same chart security-posture pass as #383. I cherry-picked it (preserving his authorship and sign-off), validated it, and added a follow-up docs commit (below).

### Breaking change (chart `0.5.0` → `0.6.0`)

This changes the default runtime posture of every workload. Overrides for the shapes that need to relax the defaults are all handled and documented in `docs/engines/k8s.md`:
- **infiniband-k8s** — the broker reads node-local fabric state and must run privileged; both shipped IB examples (`values.k8s.ib-example.yaml`, `values.slinky.ib.block-example.yaml`) set a **complete** privileged override (`privileged: true` **with** `allowPrivilegeEscalation: true`), avoiding a Helm deep-merge that would otherwise render an admission-invalid pod.
- **slurm / graph engines** — write `topology.conf` to the container filesystem, so they set `readOnlyRootFilesystem: false` plus a writable volume.
- **enforcing `restricted` on the namespace** — the docs now also cover the two shapes that are rejected when the namespace itself enforces `restricted`: the privileged IB broker (isolate it / don't enforce `restricted` on the Topograph namespace when using IB) and user-supplied `initContainers` (must carry their own `restricted`-compliant `securityContext`).

The CHANGELOG entry is marked BREAKING with a migration table.

### Validation

- `make chart-test` — 98/98, snapshots consistent (new `subchart_securitycontext_test.yaml`; a per-key-override-wins test proves the deep-merge).
- **Rendered** the default and both IB examples: default hardened on all three workloads; both IB examples produce a *valid* privileged override.
[...]

PR #410, opened 2026-07-17
Add opt-in ValidatingAdmissionPolicy for node-data-broker

### LINKED ISSUE #388: RBAC: opt-in ValidatingAdmissionPolicy to scope node-data-broker node writes (audit F1)
## Problem

The `node-data-broker` is a **DaemonSet** (`charts/topograph/charts/node-data-broker/templates/daemonset.yaml:6`) — one pod per node, each mounting the same ServiceAccount token. Its ClusterRole grants `nodes: [get,list,update]` cluster-wide whenever the subchart is enabled (`charts/topograph/charts/node-data-broker/templates/rbac.yaml:9-11`).

Kubernetes RBAC cannot scope this to "only my own node": `resourceNames` is a static list and cannot template per-pod, and the `NodeRestriction` admission plugin constrains only the `system:nodes` group (kubelet client certs), not ServiceAccount tokens. So any single compromised node — or any workload that obtains a broker pod's token — can `update` **every** Node object in the cluster. Because the schedulers that consume topograph's output (KAI, Kueue TAS, Slinky/Slurm) trust the topology data on nodes, a cluster-wide node-write is the largest blast radius in the chart.

A CEL `ValidatingAdmissionPolicy` (VAP) is the only mechanism that can constrain what the broker's SA is allowed to change on a Node.

## Evidence (re-verified against the repo, not the audit report)

The broker's entire node-write surface is **annotations under the `topograph.nvidia.com/` prefix** — nothing else:

- Write path: `cmd/node-data-broker/main.go:176` Gets its own node (`b.nodeName`), `:181` merges annotations via `mergeNodeAnnotations`, `:183` Updates. `mergeNodeAnnotations` (`cmd/node-data-broker/main.go:270-275`) touches **only** `node.Annotations` (`maps.Copy`) — it never writes labels and never deletes keys.
- Keys, per provider dispatched from `getAnnotations` (`cmd/node-data-broker/main.go:247-268`), all defined in `pkg/topology/topology.go:43-45`:

| Key | Constant | Value | Written by |
|---|---|---|---|
| instance | `KeyNodeInstance` | `topograph.nvidia.com/instance` | aws `imds.go:89`, gcp `imds.go:89`, oci `imds.go:122`, nebius `imds.go:79`, dra `provider.go:130`, lambdai `k8s.go:55`, infiniband `k8s.go:129` |
| region | `KeyNodeRegion` | `topograph.nvidia.com/region` | same seven providers (aws `imds.go:90` … infiniband `k8s.go:130`) |
| cluster-id | `KeyGpuClusterID` | `topograph.nvidia.com/cluster-id` | infiniband `k8s.go:140` only, and only when `!useGpuCliqueLabel` |

- The `--set`/`extraArgs` "extras" are **not** written to the node. The only provider that receives them, infiniband, uses them purely for config lookup (`pkg/providers/infiniband/k8s.go:84-94,146-148`) — they never enter the returned annotations map. So a CEL policy restricting changes to the `topograph.nvidia.com/*` keyspace does **not** break any legitimate write, including infiniband with custom `gpu-operator-namespace` / `device-plugin-daemonset` overrides.

The audit report's F1 (`localdev/rbac-audit-2026-07-09/report.md:82-92`) reaches the same conclusion; the enumeration above is independently derived from source. Note: the report's shorthand "relabel/annotate" is imprecise for the broker specifically — the broker writes only annotations, so the tightest correct policy also forbids label changes.

## Proposed approach

Ship an **opt-in, default-OFF** VAP + `ValidatingAdmissionPolicyBinding` in the node-data-broker subchart, gated on a new `node-data-broker.admissionPolicy.enabled` value **and** on the VAP API being served by the cluster. The CEL:

1. `matchConditions` restricts the policy to the broker's SA username only, so kubelet and all other identities are untouched (fail-safe: non-broker node updates never evaluate the constraints).
2. `validations` require: node **labels unchanged**; every added/modified annotation key is under `topograph.nvidia.com/`; no non-topograph annotation removed.
3. A best-effort per-node validation: when the cluster issues node-bound SA tokens (`ServiceAccountTokenNodeBinding`), the token carries `authentication.kubernetes.io/node-name` in `request.userInfo.extra`; the policy then requires the write to target that node. It is a **no-op** on clusters that don't populate the claim, so it never breaks a legitimate write.

## Breaking-change / compatibility note

- **Non-breaking, additive, default-OFF.** Existing installs are unaffected until an operator sets `admissionPolicy.enabled: true`.
- **Version floor:** VAP is GA (`admissionregistration.k8s.io/v1`) in Kubernetes **1.30**; the chart's `kubeVersion` floor is `>=1.27.0-0` (`charts/topograph/Chart.yaml:4`). The template is guarded on `.Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy"`, so on a 1.27–1.29 cluster the resources are silently skipped and `helm install` does not fail. (Consequence, mirroring the chart's existing `httproute.yaml` capability note: offline `helm template` won't render the VAP unless you pass `--api-versions "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy"`, because Helm only discovers real API versions at install time.)

## Residual risk (honest)

- **No true per-node pinning without node-bound tokens.** All broker pods share one SA identity; `request.userInfo.username` is identical across every node's pod. On clusters that do **not** issue node-bound SA tokens, a compromised broker token can still write `topograph.nvidia.com/*` annotations to **any** node — it just can no longer touch labels, other annotations, or (unaffected-identity) anything else. This shrinks the blast radius from "arbitrary write on every node" to "topograph-annotation write on every node," and adds true per-node pinning only where node-bound tokens are in effect.
- **Topology annotations remain writable on any node.** Because those annotations feed topology generation, a compromised token could still poison a node's `instance`/`region`/`cluster-id`. Containment is real but partial; it is defense-in-depth, not a complete fix.
- **Node `spec`/taints are not constrained** by this policy version (only labels + annotations). The broker never touches spec, so a spec-equality validation is a candidate hardening, but it is deferred to avoid any risk of rejecting a legitimate annotation-only update until validated live on a kind cluster.
- **`failurePolicy: Fail`** (fail-closed) means a future apiserver CEL-evaluation error on the broker's own updates would block annotation refresh; the matchCondition and validations are written to not error on well-formed requests, but this is the intended security trade-off.

## Acceptance criteria

- [ ] New template `charts/topograph/charts/node-data-broker/templates/validatingadmissionpolicy.yaml`, default-OFF, gated on `admissionPolicy.enabled` **and** the VAP API capability.
- [ ] New `node-data-broker.admissionPolicy` value block with `values.schema.json` validation; `docs/engines/k8s.md` subsection; `CHANGELOG.md` `[Unreleased]` entry.
- [ ] helm-unittest suite proving: renders (policy+binding) only when enabled **and** the VAP API is present; skipped when disabled; skipped when the API is absent even if enabled; CEL contains the broker SA username, the `topograph.nvidia.com/` keyspace restriction, the label-immutability rule, and `validationActions: [Deny]`; the SA username follows a custom `serviceAccount.name`.
- [ ] Default-values umbrella render snapshot is **unchanged** (the policy must not leak into a default render).
- [ ] `make chart-test` passes.

---

---
*A staged implementation plan (helm-unittest cases, template edits, docs, CHANGELOG) exists and can follow this issue.*

Shankar-v27:
## Description

Closes #388.

This PR adds an opt-in `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` to restrict the `node-data-broker` ServiceAccount to updating only its own Node resource.

### Changes

* Add `nodeDataBroker.validatingAdmissionPolicy.enabled` (default: `false`)
* Add schema validation for the new configuration
* Add `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` templates
* Add Helm unit tests
* Update Kubernetes documentation
* Add a changelog entry

### Testing

Added Helm unit tests covering:

* Default disabled rendering
* Enabled rendering
* Policy and binding generation
* `failurePolicy`
* `validationActions`
* CEL expression
* ServiceAccount interpolation

> **Note:** The Helm unit tests were added as part of this change. I was unable to execute them locally because the required Helm tooling was unavailable in my development environment. They will be executed as part of the repository's CI checks.


charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,34 @@
...
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["UPDATE"]
+        resources: ["nodes"]
+  validations:
+    - expression: >-
+        request.userInfo.username != 'system:serviceaccount:{{ .Release.Namespace }}:{{ include "nodeDataBroker.serviceAccountName" . }}' ||
+        (has(request.userInfo.extra) &&
+        has(request.userInfo.extra['authentication.kubernetes.io/node-name']) &&
```
ArangoGutierrez: This has() call takes an indexed map key, which CEL rejects at compile time (`invalid argument to has() macro` — has() only accepts a field selection). The policy fails to compile, so the API server refuses it on apply and the restriction never takes effect. Use `'authentication.kubernetes.io/node-name' in request.userInfo.extra` for the presence check. The helm-unittest suite only regex-matches the rendered string, so it doesn't catch this.

charts/topograph/values.yaml
```diff
@@ -312,6 +312,11 @@ nodeDataBroker:
   rbac:
     create: true
 
+  # Opt-in ValidatingAdmissionPolicy to scope node writes. Requires
+  # Kubernetes >= 1.26 (GA in 1.30) and ServiceAccountTokenNodeBinding enabled.
```
ArangoGutierrez: The template emits `admissionregistration.k8s.io/v1`, which is only served at GA in 1.30 (v1alpha1 in 1.26, v1beta1 in 1.28-1.29). With Chart.yaml kubeVersion `>=1.27.0-0`, enabling this on a 1.27-1.29 cluster fails to apply. Worth stating the real floor (>= 1.30) here and in the docs prerequisites.

docs/engines/k8s.md
```diff
@@ -302,6 +302,22 @@ Apply alongside the chart. A bundled template is under consideration.
...
+nodeDataBroker:
+  validatingAdmissionPolicy:
+    enabled: true
+```
+
+When enabled, the policy matches node `UPDATE` operations performed by the broker's ServiceAccount and validates that the name of the node being modified matches the node name claim bound to the requester's ServiceAccount token (`authentication.kubernetes.io/node-name`).
+
+**Prerequisites:**
+1. A Kubernetes cluster with the **`ValidatingAdmissionPolicy` API enabled** (`admissionregistration.k8s.io/v1`).
+2. ServiceAccount tokens carrying the node-binding identity claim (e.g., via `ServiceAccountTokenNodeBinding`). The policy denies updates from the broker ServiceAccount if this required node-name claim is missing from the request context.
```
ArangoGutierrez: For the broker's default pod-bound token, the `authentication.kubernetes.io/node-name` extra is populated by the ServiceAccountTokenPodNodeInfo feature gate, not ServiceAccountTokenNodeBinding (which mints node-bound tokens the chart doesn't request). Enabling only the named gate can leave the claim absent, and since the policy is fail-closed that denies every broker node update.

charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if and .Values.nodeDataBroker.enabled .Values.nodeDataBroker.validatingAdmissionPolicy.enabled }}
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: {{ include "nodeDataBroker.fullname" . }}
+  labels:
+    {{- include "nodeDataBroker.labels" . | nindent 4 }}
+spec:
+  failurePolicy: Fail
```
ArangoGutierrez, line 9: failurePolicy: Fail with a matchConstraints that selects every node UPDATE from any user means all node updates (including kubelet heartbeats) run through this CEL; a fault would deny them cluster-wide. Moving the SA check into spec.matchConditions (username == the broker SA) limits both the evaluation and the fail-closed blast radius to just the broker's requests.

charts/topograph/tests/node-data-broker_validatingadmissionpolicy_test.yaml
```diff
@@ -0,0 +1,101 @@
...
+      nodeDataBroker:
+        enabled: true
+        validatingAdmissionPolicy:
+          enabled: true
+        serviceAccount:
+          create: true
+          name: custom-node-data-broker-sa
+    documentIndex: 0
+    asserts:
+      - matchRegex:
```
ArangoGutierrez: These substring matchRegex checks stay green even if the name-binding equality were dropped or inverted, and helm-unittest never compiles the CEL — which is how the invalid has() above passes CI. Consider asserting the full spec.validations[0].expression (the same contains->equal tightening as #385) and validating the policy on a live cluster as #392 did.

ArangoGutierrez (commented):
Nice defense-in-depth addition. A few issues keep it from working as intended as written — most importantly the CEL doesn't compile, and the version and feature-gate prerequisites are inaccurate. Details inline.

1. This has() call takes an indexed map key, which CEL rejects at compile time (`invalid argument to has() macro` — has() only accepts a field selection). The policy fails to compile, so the API server refuses it on apply and the restriction never takes effect. Use `'authentication.kubernetes.io/node-name' in request.userInfo.extra` for the presence check. The helm-unittest suite only regex-matches the rendered string, so it doesn't catch this. (charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml:20)
2. The template emits `admissionregistration.k8s.io/v1`, which is only served at GA in 1.30 (v1alpha1 in 1.26, v1beta1 in 1.28-1.29). With Chart.yaml kubeVersion `>=1.27.0-0`, enabling this on a 1.27-1.29 cluster fails to apply. Worth stating the real floor (>= 1.30) here and in the docs prerequisites. (charts/topograph/values.yaml:316)
3. For the broker's default pod-bound token, the `authentication.kubernetes.io/node-name` extra is populated by the ServiceAccountTokenPodNodeInfo feature gate, not ServiceAccountTokenNodeBinding (which mints node-bound tokens the chart doesn't request). Enabling only the named gate can leave the claim absent, and since the policy is fail-closed that denies every broker node update. (docs/engines/k8s.md:319)
4. failurePolicy: Fail with a matchConstraints that selects every node UPDATE from any user means all node updates (including kubelet heartbeats) run through this CEL; a fault would deny them cluster-wide. Moving the SA check into spec.matchConditions (username == the broker SA) limits both the evaluation and the fail-closed blast radius to just the broker's requests. (charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml:9)
5. These substring matchRegex checks stay green even if the name-binding equality were dropped or inverted, and helm-unittest never compiles the CEL — which is how the invalid has() above passes CI. Consider asserting the full spec.validations[0].expression (the same contains->equal tightening as #385) and validating the policy on a live cluster as #392 did. (charts/topograph/tests/node-data-broker_validatingadmissionpolicy_test.yaml:90)

Shankar-v27:
Thanks for the detailed review and for taking the time to go through the implementation so thoroughly.

These are all valuable points. I'll go through each of them carefully:

* verify and correct the CEL expression if needed,
* update the documented Kubernetes version requirements and feature-gate prerequisites,
* evaluate moving the ServiceAccount check into `matchConditions`,
* and strengthen the Helm tests to better validate the generated policy.

I'll put together a follow-up commit addressing the applicable items after validating them against the Kubernetes API behavior and documentation.

charts/topograph/tests/node-data-broker_validatingadmissionpolicy_test.yaml
```diff
@@ -0,0 +1,105 @@
...
+          of: ValidatingAdmissionPolicy
+      - equal:
+          path: apiVersion
+          value: admissionregistration.k8s.io/v1
+      - equal:
+          path: metadata.name
+          value: chart-ci-topograph-node-data-broker
+      - equal:
+          path: spec.failurePolicy
+          value: Fail
```
ArangoGutierrez, line 54: Nothing in the suite asserts spec.matchConstraints. I mutated the template two ways and ran make chart-test on a clean tree for each: flipping operations to ["CREATE"], and deleting the matchConstraints block outright. Both leave all 109 tests green, so a change that widens the policy's scope or makes it a no-op on node updates would ship undetected. An equal on spec.matchConstraints.resourceRules[0] closes it, same shape as the rules assertion in node-data-broker_rbac_test.yaml.

charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml
```diff
@@ -0,0 +1,37 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["UPDATE"]
+        resources: ["nodes"]
+  matchConditions:
+    - name: match-node-data-broker-serviceaccount
+      expression: >-
+        request.userInfo.username == 'system:serviceaccount:{{ .Release.Namespace }}:{{ include "nodeDataBroker.serviceAccountName" . }}'
```
ArangoGutierrez, line 19: The resolved ServiceAccount name lands directly inside a single-quoted CEL literal, so an apostrophe in that value changes what the expression means rather than just how it reads. Given the policy is fail-closed across every node UPDATE, a DNS-1123 guard in _validation.tpl (the pattern already used for ingress and gatewayAPI) is cheap insurance. Non-blocking.

ArangoGutierrez (dismissed):
Re-reviewed at 733939a. All five findings from the previous round are fixed, and I verified each rather than taking the diff at face value: the has() call is now a field selection so the CEL compiles, matchConditions bounds the fail-closed blast radius to the broker's own ServiceAccount, and the 1.30 floor lines up with both the VAP v1 GA and ServiceAccountTokenPodNodeInfo (beta, default-on in 1.30 per pkg/features in release-1.30). NODE_NAME comes from fieldRef spec.nodeName, the same source the apiserver uses for the node-name claim, so the two can never disagree and the policy is a no-op on the happy path. The broker only ever writes its own node, and the ClusterRole grants nodes get/update with no nodes/status, so the policy covers the whole granted surface. make chart-test passes on the head commit. Approving; two non-blocking notes below.

1. Nothing in the suite asserts spec.matchConstraints. I mutated the template two ways and ran make chart-test on a clean tree for each: flipping operations to ["CREATE"], and deleting the matchConstraints block outright. Both leave all 109 tests green, so a change that widens the policy's scope or makes it a no-op on node updates would ship undetected. An equal on spec.matchConstraints.resourceRules[0] closes it, same shape as the rules assertion in node-data-broker_rbac_test.yaml. (charts/topograph/tests/node-data-broker_validatingadmissionpolicy_test.yaml:54)
2. The resolved ServiceAccount name lands directly inside a single-quoted CEL literal, so an apostrophe in that value changes what the expression means rather than just how it reads. Given the policy is fail-closed across every node UPDATE, a DNS-1123 guard in _validation.tpl (the pattern already used for ingress and gatewayAPI) is cheap insurance. Non-blocking. (charts/topograph/templates/nodeDataBroker/validatingadmissionpolicy.yaml:19)

Shankar-v27:
Thanks for the additional suggestions and for the thorough review. I appreciate the feedback. Since these are non-blocking and the feature is currently complete, I'll leave them as potential follow-up improvements unless you'd prefer they be included in this PR.

Shankar-v27:
I have resolved the merge conflict that existed in the CHANGELOG.md file.

ArangoGutierrez (commented):
Re-checked at 9af4fbe. The only commit since the last round is the main merge, so the code itself is byte-identical to what I reviewed at 733939a and the reasoning there still holds. I ran make chart-test on the merged head: 18 suites, 117 tests, 145 snapshots, all green. The merge did leave one thing behind, below.

Two notes: the two non-blocking items from the previous round (asserting spec.matchConstraints, and a DNS-1123 guard on the ServiceAccount name) are still open, which is fine if you'd rather land this as-is. Separately, that round's approval was auto-dismissed when the merge commit landed, so this needs a fresh approving review before it can merge.

1. The merge kept both the branch's copy of this entry and main's, so the Security section now lists it twice. This copy is the superseded wording - it's missing the `Slinky receives Node patch without Node update` clause that main's copy carries. Dropping this line leaves the ValidatingAdmissionPolicy entry as this PR's only CHANGELOG change, which is what it should be since nothing here touches RBAC. (CHANGELOG.md:56)

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Security
 
+- Opt-in `ValidatingAdmissionPolicy` and binding to restrict the `node-data-broker` ServiceAccount to only updating its host Node resource, mitigating privilege escalation via node spec or annotation manipulation (audit F1 compliance).
+- Removed unused RBAC verbs from the Topograph API server and node-data-broker ClusterRoles (least-privilege): API server `pods` rule dropped `get` (list-only), `daemonsets` rule dropped `list` (get-only), and the Slinky `configmaps` rule dropped `list`; node-data-broker `nodes` rule dropped `list` (get/update), and the InfiniBand `daemonsets`/`pods` rules dropped `list`/`get` respectively (get-only, list-only).
```
ArangoGutierrez, line 56: The merge kept both the branch's copy of this entry and main's, so the Security section now lists it twice. This copy is the superseded wording - it's missing the `Slinky receives Node patch without Node update` clause that main's copy carries. Dropping this line leaves the ValidatingAdmissionPolicy entry as this PR's only CHANGELOG change, which is what it should be since nothing here touches RBAC.

ArangoGutierrez (approved):
Approving at 9af4fbe. The policy is unchanged from the head I verified at 733939a, and make chart-test passes on the merged head: 18 suites, 117 tests, 145 snapshots.

Taking the three open items as non-blocking. The duplicated Security entry at CHANGELOG.md:56 will ship to main as-is on a squash merge, so it wants a follow-up, along with the spec.matchConstraints assertion and the DNS-1123 guard on the ServiceAccount name.

ArangoGutierrez:
@dmitsh approved and green on my side — opt-in ValidatingAdmissionPolicy scoping the node-data-broker to its own Node. I re-verified the CEL, the matchConditions blast-radius bound, and the 1.30 floor, and make chart-test passes on the merged head. Ready for merge whenever you are. Three non-blocking follow-ups are noted inline, including a duplicated CHANGELOG Security entry the main merge left behind.

dmitsh:
> @dmitsh approved and green on my side 

Thank you @Shankar-v27, @ArangoGutierrez !
Merging

PR #411, opened 2026-07-19
feat(topology): support variable network fabric tiers

### LINKED ISSUE #408: [BUG]: Kubernetes engine leaves stale higher-tier topology labels when the network hierarchy shrinks
### Version

v0.5.0

### Which installation method(s) does this occur on?

Source

### Describe the bug.

When Topograph uses the `k8s` engine and regenerates topology after the network hierarchy becomes shallower, labels that represented tiers which no longer exist remain on the Kubernetes node.

For example, suppose a node was previously labeled with:

```text
network.topology.nvidia.com/leaf=leaf-a
network.topology.nvidia.com/spine=spine-a
network.topology.nvidia.com/core=core-a
```

If a later topology graph contains only a leaf tier for that node, Topograph updates the leaf label but does not remove the old `spine` and `core` labels. Consumers therefore see topology metadata that no longer matches the provider's current graph, which can result in incorrect topology-aware scheduling or observability decisions.

The current implementation appears to explain the behavior:

- [`pkg/engines/k8s/labeler.go`](https://github.com/NVIDIA/topograph/blob/v0.5.0/pkg/engines/k8s/labeler.go#L65-L92) builds a label map containing only the tiers present in the current `topology.Graph`.
- [`pkg/engines/k8s/kubernetes.go`](https://github.com/NVIDIA/topograph/blob/v0.5.0/pkg/engines/k8s/kubernetes.go#L71-L92) then calls `MergeNodeLabels`, which uses `maps.Copy(node.Labels, labels)`.
- `maps.Copy` only adds or overwrites entries; it does not delete configured Topograph-managed labels that are absent from the newly generated map.

**Expected behavior:** each regeneration should reconcile the configured Topograph-managed topology labels with the current graph. For a node included in the new graph, obsolete higher-tier labels should be removed while unrelated node labels remain untouched. Existing `nvidia.com/gpu.clique` handling should also remain unchanged.

### Minimum reproducible example

```shell
# 1. Configure Topograph with engine: k8s.
# 2. Generate a topology in which <node-name> has leaf, spine, and core tiers.
kubectl get node <node-name> -o json | \
  jq '.metadata.labels | with_entries(select(.key | startswith("network.topology.nvidia.com/")))'

# 3. Change the provider topology/model so the same node has only a leaf tier.
# 4. Trigger topology generation again with the updated provider payload.
id=$(curl -s -X POST -H "Content-Type: application/json" \
  -d @payload.json http://localhost:49021/v1/generate)
curl -s "http://localhost:49021/v1/topology?uid=$id"

# 5. Inspect the node again.
kubectl get node <node-name> -o json | \
  jq '.metadata.labels | with_entries(select(.key | startswith("network.topology.nvidia.com/")))'

# Actual: leaf is refreshed, but the previous spine/core labels remain.
# Expected: stale spine/core labels are removed.
```

### Relevant log output

```shell
N/A — the update succeeds and no error is logged; the resulting node labels are stale.
```

### Full env printout

```shell
N/A — identified from the v0.5.0 source path and reproducible at the label reconciliation layer.
```

### Other/Misc.

Suggested regression coverage:

- A node initially has leaf/spine/core; a subsequent update contains leaf only; spine/core are removed.
- Unrelated labels on the node are preserved.
- Configured custom keys from `topologyNodeLabels` are reconciled rather than relying only on the default keys.
- Existing `nvidia.com/gpu.clique` behavior remains unchanged.

A focused unit-level reproduction is also possible by starting with a `corev1.Node` whose labels contain leaf/spine/core, calling `MergeNodeLabels` with an input map containing only leaf, and observing that spine/core remain.

Related testing roadmap: #263 proposes a topology-change E2E suite that asserts there is no stale label leak.

### Code of Conduct

- [x] I agree to follow __THIS PROJECT__'s Code of Conduct
- [x] I have searched the [open bugs](https://github.com/jarmak-nv/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

### LINKED ISSUE #412: [FEA] Support variable-depth fabric tiers
### Description

Topograph currently models the fabric topology using the fixed roles leaf,
spine, and core. This does not represent environments with arbitrary fabric
depth.

We propose replacing these fixed fabric roles with a variable-length
`FabricTiers` hierarchy. Fabric switches are ordered closest to the compute node
first, with no fixed maximum number of tiers.

Kubernetes fabric topology labels would use indexed keys:

`network.topology.nvidia.com/tier-N`

Tier 0 is the fabric tier closest to the node, and higher numbers identify each
successive tier outward.

Accelerator topology remains a single dimension represented by `AcceleratorID`
and `Graph.Domains`. Its default Kubernetes label remains:

`network.topology.nvidia.com/accelerator`

Users can fully customize the label keys through the Kubernetes engine
configuration. `fabricLabels` is a closest-first array, while `acceleratorLabel`
configures the single accelerator label. This makes it possible to generate the
previous accelerator, leaf, spine, and core label set when compatibility with
existing consumers is required.

The NFD engine does not require custom label configuration because it does not
write Kubernetes node labels. It should expose corresponding attributes using:

- `fabric-tier-N`
- `accelerator`

This change affects the canonical fabric topology representation and therefore
requires coordinated updates across providers, engines, simulation models,
tests, Helm configuration, and documentation.

dmitsh:
Replace the fixed leaf, spine, and core topology model with variable-depth, closest-first fabric tiers.
This change:
Replaces ToThreeTierGraph with arbitrary-depth ToGraph.

Represents per-instance fabric topology using:
- FabricTiers []FabricTier

Uses tier 0 for the switch closest to the compute node, increasing outward.

Publishes Kubernetes topology labels as:
- network.topology.nvidia.com/tier-N
- network.topology.nvidia.com/accelerator

Updates the Kubernetes and NFD engines to emit any number of discovered fabric tiers.

Retains the single AcceleratorID and Graph.Domains representation for Slurm block-topology compatibility.

Configures label names through Kubernetes engine parameters:
- fabricLabels
- acceleratorLabel

Allows users to reproduce the previous accelerator, leaf, spine, and core label set through custom label configuration.

Prevents mixed-depth paths sharing an outer switch ID from overwriting one another.

Updates providers, simulation models, documentation, Helm configuration, tests, and release notes.

This is a breaking change for consumers using the previous fixed leaf, spine, and core Kubernetes label keys or the process-wide Helm/CLI label overrides.

closes #412
closes #408


docs/design/nfd-engine-sdd.md
```diff
@@ -20,10 +20,10 @@ NFD CRs for consumers that already watch NFD.
 
 Topograph already maps topology into four Kubernetes label dimensions:
 
-- `network.topology.nvidia.com/accelerator`
-- `network.topology.nvidia.com/leaf`
-- `network.topology.nvidia.com/spine`
-- `network.topology.nvidia.com/core`
+- `accelerated.topology.nvidia.com/domain-0`
```
ArangoGutierrez: Why `accelerated` and not accelerator or gpu?

pkg/engines/k8s/kubernetes.go
```diff
@@ -46,32 +36,62 @@ func (eng *K8sEngine) AddNodeLabels(ctx context.Context, nodeName string, labels
...
+		if key == topology.KeyNvidiaGPUClique {
+			continue
+		}
+		if isManagedLevelLabel(key, keys) {
+			delete(labels, key)
+		}
+	}
+}
+
+func isManagedLevelLabel(key string, keys *TopologyLabelKeys) bool {
```
ArangoGutierrez, line 67: isManagedLevelLabel only recognizes the configured keys and the new tier-N/domain-N prefixes, so the retired accelerator/leaf/spine/core labels are never removed from nodes labeled by a pre-upgrade Topograph -- the legacy-core case in kubernetes_test.go pins that behavior in. Since KAI Scheduler, NVSentinel, and Kueue consume those keys, frozen pre-upgrade values could silently mislead them; special-casing the four legacy defaults for removal (or documenting manual cleanup in the changelog) would close the gap.

pkg/engines/nfd/engine.go
```diff
@@ -120,7 +120,9 @@ func (eng *NfdEngine) GenerateOutput(ctx context.Context, graph *topology.Graph,
 		}
 	}
 
-	nodeLabels, err := k8sengine.NewTopologyLabeler().BuildNodeLabels(graph)
+	nodeLabels, err := k8sengine.NewTopologyLabeler(
+		k8sengine.NewTopologyLabelKeys(nil, nil),
```
ArangoGutierrez: Hardcoding NewTopologyLabelKeys(nil, nil) drops the operator key customization the NFD engine honored via CurrentTopologyLabelKeys() until #402, and nfd Params has no fabricLabels/acceleratedLabels equivalent -- customized keys silently revert to defaults after upgrade. Plumbing the same params through the nfd engine (or calling the loss out in the changelog) would make the break explicit.

pkg/engines/nfd/objects.go
```diff
@@ -95,31 +84,30 @@ func buildNFDObjects(nodeLabels k8sengine.NodeLabelMap, gpuCliqueValues map[stri
...
-			if _, ok := groupValues[label.kind]; !ok {
-				groupValues[label.kind] = make(map[string]string)
+			elements[kind] = value
+			if _, ok := groupValues[kind]; !ok {
+				groupValues[kind] = make(map[string]string)
 			}
-			groupValues[label.kind][value] = labelKey
+			groupValues[kind][value] = labelKey
+		}
+		if gpuCliqueValue := strings.TrimSpace(gpuCliqueValues[nodeName]); gpuCliqueValue != "" {
```
ArangoGutierrez: When a node carries both a provider-derived accelerated-domain-0 value and a differing nvidia.com/gpu.clique label, the first pass has already recorded the provider value in groupValues and this clique branch only overwrites elements -- producing a NodeFeatureGroup that no NodeFeature ever matches, which cleanupResource never removes because its name stays in objectNames. Resolving the clique override before recording groupValues (as the pre-refactor single pass did) avoids the dangling CR.

ArangoGutierrez (commented):
Found 5 issues.

1. isManagedLevelLabel only recognizes the configured keys and the new tier-N/domain-N prefixes, so the retired accelerator/leaf/spine/core labels are never removed from nodes labeled by a pre-upgrade Topograph -- the legacy-core case in kubernetes_test.go pins that behavior in. Since KAI Scheduler, NVSentinel, and Kueue consume those keys, frozen pre-upgrade values could silently mislead them; special-casing the four legacy defaults for removal (or documenting manual cleanup in the changelog) would close the gap. (pkg/engines/k8s/kubernetes.go:67)
2. Hardcoding NewTopologyLabelKeys(nil, nil) drops the operator key customization the NFD engine honored via CurrentTopologyLabelKeys() until #402, and nfd Params has no fabricLabels/acceleratedLabels equivalent -- customized keys silently revert to defaults after upgrade. Plumbing the same params through the nfd engine (or calling the loss out in the changelog) would make the break explicit. (pkg/engines/nfd/engine.go:124)
3. When a node carries both a provider-derived accelerated-domain-0 value and a differing nvidia.com/gpu.clique label, the first pass has already recorded the provider value in groupValues and this clique branch only overwrites elements -- producing a NodeFeatureGroup that no NodeFeature ever matches, which cleanupResource never removes because its name stays in objectNames. Resolving the clique override before recording groupValues (as the pre-refactor single pass did) avoids the dangling CR. (pkg/engines/nfd/objects.go:104)
4. This reshapes the canonical Graph (Domains -> AcceleratedDomains) and retires the four fixed label keys, but no issue is linked from the PR; the contribution docs gate pkg/topology changes on prior issue discussion. Linking the discussion (or opening one) would satisfy that gate. (pkg/topology/topology.go:64)
5. The label families this PR adds are tier-N and domain-N; there is no level-N label, so following this line to grep node labels matches nothing. Same for the accelerated level-0 phrasing here. (docs/providers/lambdai.md:183)

pkg/topology/topology.go
```diff
@@ -65,16 +56,25 @@ const (
...
-// Tiers is the root of the switch hierarchy. Domains maps accelerator/block
-// domains to hosts and carries the finalized enumerated domain ID.
+// Tiers is the root of the switch hierarchy.
 type Graph struct {
-	Tiers   *Vertex
-	Domains DomainMap
+	Tiers *Vertex
+	// AcceleratedDomains contains accelerated-network domain maps ordered from
+	// the narrowest domain outwards. Domain zero is the topology/block source.
+	AcceleratedDomains []DomainMap
```
ArangoGutierrez: This reshapes the canonical Graph (Domains -> AcceleratedDomains) and retires the four fixed label keys, but no issue is linked from the PR; the contribution docs gate pkg/topology changes on prior issue discussion. Linking the discussion (or opening one) would satisfy that gate.

docs/providers/lambdai.md
```diff
@@ -180,7 +180,7 @@ helm install topograph oci://ghcr.io/nvidia/topograph/topograph \
 kubectl get nodes --show-labels | grep network.topology.nvidia
 ```
 
-Only `leaf`/`spine`/`core` labels appear until the API returns `nvlink` data (see the note above); the `accelerator` label follows once it does.
+Only fabric `level-N` labels appear until the API returns `nvlink` data (see the note above); accelerated `level-0` follows once it does.
```
ArangoGutierrez: The label families this PR adds are tier-N and domain-N; there is no level-N label, so following this line to grep node labels matches nothing. Same for the accelerated level-0 phrasing here.

pkg/engines/k8s/kubernetes.go
dmitsh: That's fine, Topograph allows you to configure label names, so it will be transparent for KAI and NVSentinel.

pkg/topology/topology.go
dmitsh: Created https://github.com/NVIDIA/topograph/issues/412

pkg/engines/nfd/objects.go
```diff
@@ -95,31 +84,30 @@ func buildNFDObjects(nodeLabels k8sengine.NodeLabelMap, gpuCliqueValues map[stri
...
+			kind := topologyTypeAccelerator
+			elements[kind] = gpuCliqueValue
+			if _, ok := groupValues[kind]; !ok {
+				groupValues[kind] = make(map[string]string)
+			}
+			groupValues[kind][gpuCliqueValue] = topology.KeyNvidiaGPUClique
```
ArangoGutierrez, line 114: When every member of an accelerator domain carries gpu.clique, the provider value recorded into groupValues in the first pass is never removed, so a NodeFeatureGroup gets emitted that no NodeFeature publishes and cleanup keeps it forever. Mirroring the k8s engine's skip-when-clique guard would close it - same point as my comment on the previous revision.

pkg/engines/k8s/kubernetes.go
```diff
@@ -46,32 +36,65 @@ func (eng *K8sEngine) AddNodeLabels(ctx context.Context, nodeName string, labels
...
```
ArangoGutierrez, line 67: Legacy leaf/spine/core keys aren't recognized here, so upgraded nodes keep them frozen alongside the new tier-N labels - consumers still keyed on the old names read stale topology, and Case 4 in the test pins that behavior. Configurability covers new deployments, not values already on nodes; either strip the three well-known legacy keys during relabel or add a manual-cleanup step to the CHANGELOG migration note.

pkg/engines/k8s/kubernetes_test.go
```diff
@@ -109,28 +107,31 @@ func TestMergeNodeLabels(t *testing.T) {
...
+				topology.FabricTierKey(1):          "new-spine",
+				"network.topology.nvidia.com/core": "legacy-core",
+				"workload.example/label":           "keep",
 			},
 		},
 		{
-			name:             "Case 5: do not overwrite GPU clique when it is the configured accelerator label",
-			acceleratorLabel: topology.KeyNvidiaGPUClique,
+			name:             "Case 5: custom accelerator label still protects GPU clique",
+			acceleratorLabel: "custom.example/accelerator",
```
ArangoGutierrez, line 153: Swapping this case away from acceleratorLabel == nvidia.com/gpu.clique removes the only coverage of the clique-preservation branches; either guard could now be deleted without a failing test, and the config is operator-reachable. Worth keeping the custom-label case and re-adding a clique case alongside.

pkg/engines/nfd/engine.go
```diff
@@ -120,7 +120,9 @@ func (eng *NfdEngine) GenerateOutput(ctx context.Context, graph *topology.Graph,
 		}
 	}
 
-	nodeLabels, err := k8sengine.NewTopologyLabeler().BuildNodeLabels(graph)
+	nodeLabels, err := k8sengine.NewTopologyLabeler(
+		k8sengine.NewTopologyLabelKeys(nil, ""),
```
ArangoGutierrez, line 124: Still hardcoded to the defaults, so the new fabricLabels/acceleratorLabel customization stops at the k8s engine - NFD operators can't use it and the limitation isn't documented. Plumbing the params through or a line in the CHANGELOG would both settle it; re-raising from the previous revision.

ArangoGutierrez (commented):
Found 4 issues.

1. When every member of an accelerator domain carries gpu.clique, the provider value recorded into groupValues in the first pass is never removed, so a NodeFeatureGroup gets emitted that no NodeFeature publishes and cleanup keeps it forever. Mirroring the k8s engine's skip-when-clique guard would close it - same point as my comment on the previous revision. (pkg/engines/nfd/objects.go:110)
2. Legacy leaf/spine/core keys aren't recognized here, so upgraded nodes keep them frozen alongside the new tier-N labels - consumers still keyed on the old names read stale topology, and Case 4 in the test pins that behavior. Configurability covers new deployments, not values already on nodes; either strip the three well-known legacy keys during relabel or add a manual-cleanup step to the CHANGELOG migration note. (pkg/engines/k8s/kubernetes.go:67)
3. Swapping this case away from acceleratorLabel == nvidia.com/gpu.clique removes the only coverage of the clique-preservation branches; either guard could now be deleted without a failing test, and the config is operator-reachable. Worth keeping the custom-label case and re-adding a clique case alongside. (pkg/engines/k8s/kubernetes_test.go:134)
4. Still hardcoded to the defaults, so the new fabricLabels/acceleratorLabel customization stops at the k8s engine - NFD operators can't use it and the limitation isn't documented. Plumbing the params through or a line in the CHANGELOG would both settle it; re-raising from the previous revision. (pkg/engines/nfd/engine.go:124)

pkg/engines/nfd/objects.go
dmitsh: It is no longer applicable to the current code. buildNFDObjects now resolves gpuCliqueValue first and skips the provider accelerator label before adding it to either elements or groupValues

pkg/engines/k8s/kubernetes.go
dmitsh: we don't have any production deployment that uses these labels

pkg/engines/nfd/engine.go
dmitsh: customized labels are only applicable to k8s engine. The nfd engine uses defaults.

Issue #415, opened 2026-07-21
Dual-level accelerated domain support


dmitsh:
### Description

Topograph currently models accelerator topology as a single locality level. Add support for two nested levels so schedulers can make placement decisions within both a base domain and its parent domain.

#### Slurm requirements
- Generate "BlockSizes=N,R*N" where N is the number of nodes in a rack and R is the number of racks in parent domain.
- Emit exactly R consecutive base blocks for each parent domain.
- Never interleave blocks from different parent domains.
- Reserve all R positions when a parent domain is first discovered.
- Represent unavailable or undiscovered base domains with empty entries so later blocks do not shift.
- Use deterministic names that identify the parent domain and base-block position.
- Apply this behavior to every Slurm topology/block output, including Slinky where applicable.

Example:
```
BlockName=domain_d1_b04 Nodes=node[028-036]
BlockName=domain_d1_b05
BlockName=domain_d1_b06 Nodes=node[046-054]
```
Slurm derives each aggregate from consecutive base blocks, making stable ordering and placeholders necessary.

Before implementation, confirm that base domains have persistent identifiers or positions. Otherwise, stable ordering cannot be guaranteed.

#### Kubernetes requirements
Expose both hierarchy levels as configurable topology labels. Label values must remain stable and be shared by all nodes in the corresponding domain.

#### Design consideration
The canonical graph currently supports only one accelerator locality dimension through `AcceleratorID` and `Graph.Domains`. Supporting nested domains may require a coordinated canonical-model change that should be designed and discussed before implementation.

PR #416, opened 2026-07-22
feat: Dual-level accelerated domain support - Design and Prototype

ravisoundar:
## Description
Design and Prototype for the dual level accelerated domain support.
Addresses #415 

## Checklist


docs/design/nvl576-multi-level-topology-sdd.md
```diff
@@ -0,0 +1,278 @@
...
+`pkg/topology/domain.go` — three new string fields:
+
+```go
+type HostInfo struct {
+    Domain     string // accelerator/NVLink domain name (rack)
+    InstanceID string
+    HostName   string
+    Level1     string // optional: top-level grouping (e.g. datacenter)
+    Level2     string // optional: mid-level grouping (e.g. building or SU group)
+    Level3     string // optional: lowest named level above the domain (e.g. SU / room)
```
dmitsh: Are you mixing network fabric and accelerated domains?
"Level-N" isn't very descriptive. It may mean different things to different providers.
Also, I'd use map[string]string instead of a fixed number of levels. The key could resemble a k8s label name, which we could then use in the k8s engine.

dmitsh (commented):
## Correctness Issues

### 1. SDD documents the wrong label name

`docs/design/dual-level-topology-sdd.md` (Data Model / Simulation model YAML section) states:

```
network.topology.nvidia.com/group → HostInfo.SubDomain (parent domain)
```

But `pkg/models/model.go` defines:

```go
LabelTopologySubDomain = "network.topology.nvidia.com/sub-domain"
```

The SDD name (`/group`) and the implementation name (`/sub-domain`) don't match. Providers reading the SDD to opt in will use the wrong label.

---

### 2. `convert`: nil return value not guarded (latent panic)

In the interior-node path of `convert` (`pkg/translate/block_tree.go`):

```go
for _, name := range slices.Sorted(maps.Keys(src.Children)) {
    converted := convert(src.Children[name], baseBlockSize)
    target.children = append(target.children, converted)
    target.nodeCount += converted.nodeCount  // panics if converted == nil
```

`convert` returns `nil` when `src == nil`. `src.Children[name]` can't be nil today (all values are set by `GetDomainTree`), but this is a latent panic for any future caller. A guard `if converted == nil { continue }` eliminates the footgun at zero cost.

---

### 3. `childCapacity` can stay zero, silently suppressing padding

```go
childCapacity := 0
for _, name := range slices.Sorted(maps.Keys(src.Children)) {
    converted := convert(src.Children[name], baseBlockSize)
    ...
    if childCapacity == 0 {
        childCapacity = converted.nodeCount
    }
}

for target.nodeCount < src.DesiredNodeCount && childCapacity > 0 {
    ...
}
```

`childCapacity` is sampled only from the first child. If that child has `nodeCount == 0` (e.g. its `DesiredNodeCount` was never set because `blockSizes` was empty, or it's a leaf domain with zero hosts), `childCapacity` stays 0 and the padding loop never fires. Absent base domains are silently omitted from the output rather than getting placeholder slots. The fix is to sample `childCapacity` from the first child with `nodeCount > 0`, not just the first child unconditionally.

---

### 4. `domainsForBlocks` interaction with `SubDomain` (pre-existing, but worth noting)

`complementBlocks` calls `domainsForBlocks(nt.domains, blocks)` which filters `nt.domains` by matching `b.name` (block name) against domain map keys. In the single-level case `b.name` is the accelerator domain name, so the lookup works. In the dual-level case `b.name` will be the *parent* domain name (`domain-01`, etc.) which IS the key in `DomainMap`, so it still works — but only because the test fixture uses the accelerator domain as the outer key.

If a future provider populates the block name as the *sub-domain* (base rack name) instead of the parent domain, `all[b.name]` will return nil and hosts will be silently dropped. The pairing between `blockInfo.name` and `DomainMap` keys is an undocumented contract worth making explicit (or enforcing in tests).

---

### 5. Placeholder positions diverge from alphabetical slot positions

`rack-1-03` and `rack-1-13` are alphabetically positions 3 and 13 within domain-01, but their placeholder blocks appear as `block015` and `block016` (the last two slots). The SDD documents this as a known limitation, but it means Slurm's position-based aggregate inference will be wrong for those specific slots — a real operational impact. The limitation callout in the SDD should be elevated to the PR description so reviewers and operators are not surprised.

pkg/models/model.go
```diff
@@ -30,8 +30,9 @@ import (
 )
 
 const (
-	LabelTopologyRegion = "topology.kubernetes.io/region"
-	LabelTopologyZone   = "topology.kubernetes.io/zone"
+	LabelTopologyRegion    = "topology.kubernetes.io/region"
+	LabelTopologyZone      = "topology.kubernetes.io/zone"
+	LabelTopologySubDomain = "network.topology.nvidia.com/sub-domain"
```
dmitsh: This label name "network.topology.nvidia.com/sub-domain" is vague. There are many domains. Which one does it define?
Maybe "accelerator.topology.nvidia.com/subdomain" will be mode descriptive.
And, we could also rename ""network.topology.nvidia.com/accelerator" to ""accelerator.topology.nvidia.com/domain" for consistency.

pkg/translate/block_tree.go
```diff
@@ -213,159 +223,97 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
+	hosts := src.Hosts()
+	desiredNodeCount := src.DesiredNodeCount()
+
+	// Leaf node: a non-nil hosts map signals "leaf" regardless of population.
+	// GetDomainTree initialises the map only for leaf vertices; interior vertices
+	// always have a nil map. Using != nil (rather than len > 0) means a fully-
+	// skipped domain still emits the correct number of empty placeholder slots
+	// instead of producing a silent 0-nodeCount aggregate via the interior path.
+	if hosts != nil {
+		groupSize := desiredNodeCount / baseBlockSize
```
ArangoGutierrez: Single-level domains now pad to `desiredNodeCount / baseBlockSize` base blocks, where `desiredNodeCount` is the smallest configured BlockSize >= the largest domain. Whenever `blockSizes[0] < maxDomainHosts < blockSizes[last]`, every domain gets padded out to a full top-tier block. A/B against the merge base on identical no-SubDomain DomainMaps: 8 domains x 18 hosts with BlockSizes=9,144 gives 16 blocks / 0 empty before and 128 / 112 after; 6 x 16 with 8,64 goes 16/4 to 48/36; 3 x 72 with 18,144 goes 16/4 to 24/12. The SDD goal states no-SubDomain domains must produce output identical to the pre-change behavior.

pkg/translate/block_tree.go
```diff
@@ -213,159 +223,97 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
 	}
-	return aggregateNodes, total
-}
 
-func newAggregateBlock(children []blockTreeNode, size int, blockSizes []int) blockTreeNode {
-	if size <= 0 {
-		return nil
+	// Pad with empty sibling aggregates until DesiredNodeCount is reached.
+	padded := 0
+	for target.nodeCount < desiredNodeCount && childCapacity > 0 {
```
ArangoGutierrez: The padding loop stops as soon as `target.nodeCount` reaches `desiredNodeCount` instead of rounding up to a multiple of it, so an incomplete top-tier group is no longer complemented. Same A/B, opposite direction: 5 single-level domains of 3 hosts with BlockSizes=4,16 emits 8 blocks before (5 real + 3 placeholders completing the second 16-node aggregate) and 5 after; 9 domains goes 12 to 9. The removed `packAggregateNodes`/`newAggregateBlock` padded every aggregate to a full groupSize, which is what protects Slurm's position-based aggregate inference.

pkg/topology/domain.go
```diff
@@ -64,3 +111,128 @@ func (m DomainMap) AddHostInfo(hostInfo *HostInfo) {
...
+			domainBV.Vertices = make(map[string]*Vertex)
+			placed := 0
+			for _, host := range hosts {
+				gn := host.SubDomain
+				if gn == "" {
+					// A host with no SubDomain in a domain where other hosts carry
+					// SubDomains indicates a partially-configured provider. Bucketing
+					// it under key "" would create a vertex that sorts before all real
+					// sub-domains, shifting block numbers and emitting a nameless block.
+					klog.Warningf("domain %q: host %q has no SubDomain while other hosts in the domain do; skipping", domain, host.HostName)
```
ArangoGutierrez: A host with an empty SubDomain inside a domain where others carry one is skipped entirely, so it never reaches a block and vanishes from topology.conf with only a klog warning. Verified: 8 racked hosts plus 1 without a SubDomain emits `Nodes=node[0001-0008]` and drops the ninth. OCI's `convertComputeHost` leaves `rack` empty whenever `additionalData.locationDetails.rack` is absent, so a partial `GetComputeHost` response silently removes those compute nodes. Before this change such a host landed in its own single-level domain and was always emitted.

pkg/engines/k8s/kubernetes.go
```diff
@@ -65,10 +65,11 @@ func removeManagedTopologyLabels(labels map[string]string, keys *TopologyLabelKe
 }
 
 func isManagedLevelLabel(key string, keys *TopologyLabelKeys) bool {
-	if key == topology.KeyTopologyAccelerator {
+	if key == topology.KeyTopologyXclrDomain || key == topology.KeyTopologyXclrSubDomain {
```
ArangoGutierrez, line 183: `isManagedLevelLabel` now matches only the new xclr keys, so `network.topology.nvidia.com/accelerator` is no longer treated as managed and `removeManagedTopologyLabels` never strips it. Nodes labelled by a previous release keep that key with a frozen value forever, alongside the new one. That string now appears nowhere in the tree except this CHANGELOG entry, so there is no cleanup path and no migration note — and CLAUDE.md names KAI Scheduler, NVSentinel and Kueue as consumers of it. Keeping the legacy key in the managed set would let one reconcile clean it up.

pkg/engines/nfd/objects.go
```diff
@@ -91,7 +92,7 @@ func buildNFDObjects(nodeLabels k8sengine.NodeLabelMap, gpuCliqueValues map[stri
 			if !ok {
 				continue
 			}
-			if kind == topologyTypeAccelerator && gpuCliqueValue != "" {
+			if kind == topologyTypeXclrDomain && gpuCliqueValue != "" {
```
ArangoGutierrez: The gpu.clique shadow check skips only `topologyTypeXclrDomain`, so when a clique is present the domain is replaced by the clique value while the provider sub-domain is still published. `TestBuildNFDObjectsGroupsXclrSubDomainsAndDoesNotGroupShadowedDomains` encodes the result: domain `gpu-clique-a` together with sub-domain `provider-domain-a.rack-a`, nested under a domain that was never published. The k8s engine's `skipXclrLabelsWhenGPUCliqueExists` drops both levels, and node-labels.md says the sub-domain is written only when the domain is. Adding `topologyTypeXclrSubDomain` to the same `continue` would line them up.

pkg/topology/domain.go
```diff
@@ -64,3 +111,128 @@ func (m DomainMap) AddHostInfo(hostInfo *HostInfo) {
...
+//     that holds its hosts directly (one level below root).
+//   - When hosts carry a SubDomain, the domain vertex has one child per distinct
+//     SubDomain value, and each sub-domain vertex holds the hosts belonging to it
+//     (two levels below root).
+//
+// DesiredNodeCount is then set on every vertex via a BFS pass: all vertices at
+// the same tree depth receive the smallest blockSize >= the maximum
+// actualNodeCount at that depth. Root (actualNodeCount == 0) always receives
+// blockSizes[last].
+func (m DomainMap) GetDomainTree(blockSizes []int) *BlockVertex {
```
ArangoGutierrez: `GetDomainTree`, `setDesiredCountByLevel` and `BlockVertex` add roughly 120 lines of exported behavior to the load-bearing topology package with no direct unit test — no `*_test.go` in the tree references any of the three, and `domain_test.go` still holds only `TestDomainMapAddHost`. The BFS level sizing, the empty-SubDomain skip path and the documented "blockSizes need not be sorted" contract are reached only indirectly through translate's rendered-output assertion, and the SDD test plan lists them as covered by pkg/topology tests.

ArangoGutierrez (commented):
Solid docs discipline here — every breaking change is in the CHANGELOG and AGENTS.md/.claude/CLAUDE.md are updated in the same PR. The main concern is the block-tree rewrite: running the old and new generators side by side on identical single-level DomainMaps, the output diverges in both directions, which the SDD says should stay identical.

1. Single-level domains now pad to `desiredNodeCount / baseBlockSize` base blocks, where `desiredNodeCount` is the smallest configured BlockSize >= the largest domain. Whenever `blockSizes[0] < maxDomainHosts < blockSizes[last]`, every domain gets padded out to a full top-tier block. A/B against the merge base on identical no-SubDomain DomainMaps: 8 domains x 18 hosts with BlockSizes=9,144 gives 16 blocks / 0 empty before and 128 / 112 after; 6 x 16 with 8,64 goes 16/4 to 48/36; 3 x 72 with 18,144 goes 16/4 to 24/12. The SDD goal states no-SubDomain domains must produce output identical to the pre-change behavior. (pkg/translate/block_tree.go:257)
2. The padding loop stops as soon as `target.nodeCount` reaches `desiredNodeCount` instead of rounding up to a multiple of it, so an incomplete top-tier group is no longer complemented. Same A/B, opposite direction: 5 single-level domains of 3 hosts with BlockSizes=4,16 emits 8 blocks before (5 real + 3 placeholders completing the second 16-node aggregate) and 5 after; 9 domains goes 12 to 9. The removed `packAggregateNodes`/`newAggregateBlock` padded every aggregate to a full groupSize, which is what protects Slurm's position-based aggregate inference. (pkg/translate/block_tree.go:287)
3. A host with an empty SubDomain inside a domain where others carry one is skipped entirely, so it never reaches a block and vanishes from topology.conf with only a klog warning. Verified: 8 racked hosts plus 1 without a SubDomain emits `Nodes=node[0001-0008]` and drops the ninth. OCI's `convertComputeHost` leaves `rack` empty whenever `additionalData.locationDetails.rack` is absent, so a partial `GetComputeHost` response silently removes those compute nodes. Before this change such a host landed in its own single-level domain and was always emitted. (pkg/topology/domain.go:161)
4. `isManagedLevelLabel` now matches only the new xclr keys, so `network.topology.nvidia.com/accelerator` is no longer treated as managed and `removeManagedTopologyLabels` never strips it. Nodes labelled by a previous release keep that key with a frozen value forever, alongside the new one. That string now appears nowhere in the tree except this CHANGELOG entry, so there is no cleanup path and no migration note — and CLAUDE.md names KAI Scheduler, NVSentinel and Kueue as consumers of it. Keeping the legacy key in the managed set would let one reconcile clean it up. (pkg/engines/k8s/kubernetes.go:68)
5. The gpu.clique shadow check skips only `topologyTypeXclrDomain`, so when a clique is present the domain is replaced by the clique value while the provider sub-domain is still published. `TestBuildNFDObjectsGroupsXclrSubDomainsAndDoesNotGroupShadowedDomains` encodes the result: domain `gpu-clique-a` together with sub-domain `provider-domain-a.rack-a`, nested under a domain that was never published. The k8s engine's `skipXclrLabelsWhenGPUCliqueExists` drops both levels, and node-labels.md says the sub-domain is written only when the domain is. Adding `topologyTypeXclrSubDomain` to the same `continue` would line them up. (pkg/engines/nfd/objects.go:95)
6. `GetDomainTree`, `setDesiredCountByLevel` and `BlockVertex` add roughly 120 lines of exported behavior to the load-bearing topology package with no direct unit test — no `*_test.go` in the tree references any of the three, and `domain_test.go` still holds only `TestDomainMapAddHost`. The BFS level sizing, the empty-SubDomain skip path and the documented "blockSizes need not be sorted" contract are reached only indirectly through translate's rendered-output assertion, and the SDD test plan lists them as covered by pkg/topology tests. (pkg/topology/domain.go:127)
7. `topology.KeyTopologyAcceleratorSubDomain` never existed — `git log -S` finds it only in this PR's own commit, and the base declares just `KeyTopologyAccelerator`. `KeyTopologyXclrSubDomain` is a new constant, so this reads better under Added than as a BREAKING rename. (CHANGELOG.md:29)
8. This says retaining a provider-supplied XCLR sub-domain when `nvidia.com/gpu.clique` is present matches the k8s engine, but `skipXclrLabelsWhenGPUCliqueExists` deletes both the domain and the sub-domain. Worth correcting in whichever direction the NFD behavior gets resolved. (docs/design/nfd-engine-sdd.md:234)
9. This delete is redundant with `removeManagedTopologyLabels`, which already treats the sub-domain key as managed. Removing the line keeps pkg/engines, pkg/topology and pkg/translate green, so nothing pins the intended behavior — "Case 8: remove stale accelerator sub-domain label" exercises the removeManagedTopologyLabels path instead. (pkg/engines/k8s/kubernetes.go:102)

CHANGELOG.md
```diff
@@ -21,9 +24,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
-- **BREAKING:** Simulation models now define accelerator topology through inherited `switches[].annotations` and `blocks[].annotations` using `accelerator.topology.test/domain` and optional `accelerator.topology.test/sub-domain`, instead of the `network.topology.nvidia.com/accelerator` model label.
+- **BREAKING:** `k8s.TopologyLabelKeys.Accelerator` and `AcceleratorKey` are renamed to `XclrDomain` and `XclrDomainKey`; the struct now also exposes `XclrSubDomain` and `XclrSubDomainKey`.
+- **BREAKING:** The exported Go constant `topology.KeyTopologyAccelerator` is renamed to `topology.KeyTopologyXclrDomain`.
+- **BREAKING:** The exported Go constant `topology.KeyTopologyAcceleratorSubDomain` is renamed to `topology.KeyTopologyXclrSubDomain`.
```
ArangoGutierrez: `topology.KeyTopologyAcceleratorSubDomain` never existed — `git log -S` finds it only in this PR's own commit, and the base declares just `KeyTopologyAccelerator`. `KeyTopologyXclrSubDomain` is a new constant, so this reads better under Added than as a BREAKING rename.

docs/design/nfd-engine-sdd.md
```diff
@@ -220,12 +224,13 @@ small patches to reduce write amplification.
...
+- Unit-test graph-to-group generation across variable fabric tiers, XCLR
+  domains, and XCLR sub-domains.
 - Verify long and invalid topology values produce stable CR names.
 - Verify stale Topograph-managed groups are removed when `cleanup` is enabled.
 - Verify an empty generated object set returns an error and preserves existing
   objects when `cleanup` is enabled.
-- Verify nodes with `nvidia.com/gpu.clique` follow the same accelerator behavior
-  as the `k8s` engine.
+- Verify nodes with `nvidia.com/gpu.clique` override the XCLR domain while
+  retaining any provider-supplied XCLR sub-domain, matching the `k8s` engine.
```
ArangoGutierrez: This says retaining a provider-supplied XCLR sub-domain when `nvidia.com/gpu.clique` is present matches the k8s engine, but `skipXclrLabelsWhenGPUCliqueExists` deletes both the domain and the sub-domain. Worth correcting in whichever direction the NFD behavior gets resolved.

pkg/engines/k8s/kubernetes.go
```diff
@@ -84,18 +85,21 @@ func isManagedLevelLabel(key string, keys *TopologyLabelKeys) bool {
...
-	delete(filtered, acceleratorLabel)
+	delete(filtered, xclrDomainLabel)
+	delete(filtered, xclrSubDomainLabel)
 
-	if acceleratorLabel != topology.KeyNvidiaGPUClique {
-		delete(node.Labels, acceleratorLabel)
+	if xclrDomainLabel != topology.KeyNvidiaGPUClique {
+		delete(node.Labels, xclrDomainLabel)
 	}
+	delete(node.Labels, xclrSubDomainLabel)
```
ArangoGutierrez: This delete is redundant with `removeManagedTopologyLabels`, which already treats the sub-domain key as managed. Removing the line keeps pkg/engines, pkg/topology and pkg/translate green, so nothing pins the intended behavior — "Case 8: remove stale accelerator sub-domain label" exercises the removeManagedTopologyLabels path instead.

pkg/translate/block_tree.go
ravisoundar: Fixed the issue, and now the pre-change behavior is restored for these tests.

pkg/translate/block_tree.go
ravisoundar: Fixed the padding to include when the node count exceeds the maximum provided blockSize.

pkg/topology/domain.go
ravisoundar: Updated the logic to include these nodes in the `accelerator domain` when the `sub-domain` classification is missing.

pkg/engines/k8s/kubernetes.go
ravisoundar: Added the old label now in the check.

pkg/topology/domain.go
ravisoundar: Added more tests now.

CHANGELOG.md
ravisoundar: Done.

pkg/engines/k8s/kubernetes.go
ravisoundar: Removed it.

pkg/engines/nfd/objects.go
ravisoundar: @dmitsh, not sure if this is the desired behavior. I assume sub-domain information may still be required.

ArangoGutierrez: Yes, it's valid — though the issue is coherence of the pair rather than the sub-domain being unwanted.

`pkg/providers/oci/instance_topology.go:167` builds the value as `XclrDomainID + "." + rack`, so the sub-domain embeds the provider domain ID as its own prefix. Once the clique replaces the domain we publish `domain=gpu-clique-a` alongside `sub-domain=provider-domain-a.rack-a`, and `provider-domain-a` appears in no group — `TestBuildNFDObjectsGroupsXclrSubDomainsAndDoesNotGroupShadowedDomains` asserts that set as `{gpu-clique-a, provider-domain-b}`. A consumer grouping by sub-domain gets a NodeFeatureGroup whose parent doesn't exist.

Three places already encode "sub-domain requires domain", and NFD is the only one that doesn't:

- `pkg/topology/graph.go` `toInstance()` — sets the sub-domain label only when the domain label is non-empty (`TestToInstanceOmitsXclrSubDomainWithoutDomainLabel`)
- `pkg/engines/k8s/kubernetes.go` `skipXclrLabelsWhenGPUCliqueExists()` — deletes both
- `docs/reference/node-labels.md:25` — "Written only when `xclr.topology.nvidia.com/domain` is also present"

On the clique carrying only a domain ID: it's `<ClusterUUID>.<CliqueID>` (node-labels.md:117) — ClusterUUID is the physical NVL Domain, CliqueID the Fabric-Manager sub-domain within it. Already two levels, and never coarser than the accelerator domain.

On rack information for scheduling: `network.topology.nvidia.com/tier-N` is not suppressed by the clique — `skipXclrLabelsWhenGPUCliqueExists` removes only the two XCLR keys. node-labels.md:70 names tier-0/1/2 as the carrier for cross-rack proximity for exactly this reason. The XCLR sub-domain is provider-namespaced and not comparable across providers (netq's DomainUUID explicitly so), so it's the weaker signal here.

Adding `topologyTypeXclrSubDomain` to the same `continue` builds clean; exactly two tests in `pkg/engines/nfd` fail, both encoding the current behavior, and `pkg/engines/k8s`, `pkg/topology` and `pkg/translate` stay green.

Related: the separate comment on `docs/design/nfd-engine-sdd.md:234` is the root cause — it instructs retaining the sub-domain "matching the `k8s` engine", which the k8s engine does not do. Worth correcting in the same commit.

ravisoundar: Added `topologyTypeXclrSubDomain` to the condition to skip it and fixed the test cases.

docs/design/nfd-engine-sdd.md
ravisoundar: Updated.

pkg/translate/block_tree.go
```diff
@@ -213,159 +223,118 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
-			count += c.nodeCount
+	// Pad to the nearest positive multiple of lcm(childCapacity, desiredNodeCount).
+	// Pre-computing the target via the LCM bounds the loop to ≤ desiredNodeCount/gcd
+	// iterations for any block-size pair, including non-power-of-2-aligned ones.
+	padded := 0
+	if desiredNodeCount > 0 && childCapacity > 0 && target.nodeCount > 0 && target.nodeCount%desiredNodeCount != 0 {
+		g := blockTreeGCD(childCapacity, desiredNodeCount)
+		step := childCapacity / g * desiredNodeCount // = lcm(childCapacity, desiredNodeCount)
+		targetCount := ((target.nodeCount + step - 1) / step) * step
+		for target.nodeCount < targetCount {
```
ArangoGutierrez: The padding loop steps by `childCapacity` and stops at `>= targetCount`, so it lands on a multiple of `desiredNodeCount` only when every child has the same capacity. Dual-level domains break that: with `blockSizes=[18,144,1152]`, domain `aaa` (72 hosts, 8 sub-domains) converts to nodeCount 576 while `bbb` (144 hosts, 2 sub-domains) converts to 144, so the root goes 720 -> 1296 and stops, emitting 72 base blocks where 1296 % 1152 = 144. A sweep of 784 two-domain shapes over that list hits it in 22, and the output then declares `BlockSizes=18,144,1152` over a block list that does not tile into 1152. It needs unequal sub-domain counts per domain, so `tests/models/dual-level.yaml` does not catch it on any block-size list I tried - but that is the shape a real OCI cluster with uneven racks per fabric domain produces.

docs/engines/nfd.md
```diff
@@ -144,10 +146,10 @@ membership even when an NFD worker does not run on the node, as with simulated
 KWOK nodes. Topograph does not write `status.nodes`; NFD owns status updates.
 
 If a Kubernetes node already has `nvidia.com/gpu.clique`, the engine uses that
-label's value as the authoritative accelerator attribute instead of the value
+label's value as the authoritative `xclr-domain` attribute instead of the value
 derived from the provider graph. The matching `NodeFeatureGroup` records
-`nvidia.com/gpu.clique` as its source label key. All fabric-tier attributes
-are still published.
+`nvidia.com/gpu.clique` as its source label key. Any provider-supplied
+`xclr-sub-domain` and all fabric-tier attributes are still published.
```
ArangoGutierrez: `buildNFDObjects` skips both `xclr-domain` and `xclr-sub-domain` when `nvidia.com/gpu.clique` is present, so a provider-supplied sub-domain attribute is not published. This PR's own `TestBuildNFDObjectsSuppressesSubDomainWhenGPUCliqueExists` asserts that suppression, and `docs/reference/node-labels.md` describes it correctly, so this line is the outlier.

pkg/engines/k8s/labeler.go
```diff
@@ -53,7 +55,10 @@ func (keys *TopologyLabelKeys) Validate() error {
 			return err
 		}
 	}
-	return validate("acceleratorLabel", keys.Accelerator)
+	if err := validate("acceleratorLabel", keys.XclrDomain); err != nil {
+		return err
+	}
+	return validate("XclrSubDomain", keys.XclrSubDomain)
```
ArangoGutierrez: This location string is the Go field name, while the other two are operator-facing config keys (`fabricLabels[N]` and `acceleratorLabel`). Pointing `acceleratorLabel` at the sub-domain key yields `configured for both acceleratorLabel and XclrSubDomain`, which names a parameter operators cannot set.

pkg/topology/domain.go
```diff
@@ -64,3 +122,181 @@ func (m DomainMap) AddHostInfo(hostInfo *HostInfo) {
...
+//     SubDomain value, and each sub-domain vertex holds the hosts belonging to it
+//     (two levels below root). A host with an empty SubDomain in an otherwise
+//     dual-level domain is placed in a fallback sub-domain vertex keyed by the
+//     accelerator domain name, so the host is always emitted.
+//
+// DesiredNodeCount is then set on every vertex via a BFS pass whose behaviour
+// depends on the number of blockSizes provided:
+//
+//   - Multiple blockSizes: all vertices at the same tree depth receive the
+//     smallest blockSize >= the maximum actualNodeCount at that depth; the root
```
ArangoGutierrez: The comment says each depth receives the smallest `blockSize >= max actualNodeCount`, but `getDesiredCountByLevel` calls `pow2GroupCapacity`, which returns the smallest `2^n x base`. With `blockSizes=[18,144]` and a depth max of 20 that is 36, not 144, so the documented rule and the code disagree on a value that drives block padding. The same claim is repeated on `setDesiredCountByLevel`.

CHANGELOG.md
```diff
@@ -21,9 +25,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+- **BREAKING:** `k8s.TopologyLabelKeys.Accelerator` and `AcceleratorKey` are renamed to `XclrDomain` and `XclrDomainKey`; the struct now also exposes `XclrSubDomain` and `XclrSubDomainKey`.
+- **BREAKING:** The exported Go constant `topology.KeyTopologyAccelerator` is renamed to `topology.KeyTopologyXclrDomain`.
+- **BREAKING:** `topology.InstanceTopology` now represents accelerator locality with `XclrDomainID` and optional `XclrSubDomainID`; these replace the ambiguous `AcceleratorID` and `ParentAcceleratorID` fields.
+- **BREAKING:** The default accelerator-domain label is now `xclr.topology.nvidia.com/domain`, replacing `network.topology.nvidia.com/accelerator`. Topograph will automatically remove the old label from each node on its first reconciliation after upgrading.
+- **BREAKING:** Simulation models now define accelerator topology through inherited `switches[].annotations` and `blocks[].annotations` using `accelerator.topology.test/domain` and optional `accelerator.topology.test/sub-domain`, instead of the `xclr.topology.nvidia.com/domain` model label.
```
ArangoGutierrez: This entry records what simulation models moved away from, and that key was `network.topology.nvidia.com/accelerator` - `xclr.topology.nvidia.com/domain` did not exist when that change landed. The rename swept up a historical reference, so the entry now describes models leaving a key they never used.

pkg/topology/domain.go
```diff
@@ -27,6 +30,61 @@ type HostInfo struct {
...
+	desiredNodeCount int
+	hosts            map[string]*HostInfo // non-nil only for leaf vertices
+}
+
+// asBlockVertex recovers the *BlockVertex whose embedded Vertex field v points
+// to. Because BlockVertex embeds Vertex as its first field, Go guarantees that
+// the addresses are identical, making the cast safe — but only for *Vertex
+// values stored as &blockVertex.Vertex. Callers must never pass a *Vertex that
+// was not obtained that way; see the BlockVertex safety invariant.
+func asBlockVertex(v *Vertex) *BlockVertex {
```
ArangoGutierrez: `asBlockVertex` is sound only while every `*Vertex` in `Vertices` is `&child.Vertex`, but `Vertices` is an exported field on an exported type, so nothing outside this file enforces that. A `children map[string]*BlockVertex` field, or an unexported child map with `ChildAt` as the only reader, would let the compiler hold the invariant instead of a doc comment.

ArangoGutierrez (commented):
Found 6 issues. The first two are worth settling before merge; the rest are docs and API-surface fixes.

1. The padding loop steps by `childCapacity` and stops at `>= targetCount`, so it lands on a multiple of `desiredNodeCount` only when every child has the same capacity. Dual-level domains break that: with `blockSizes=[18,144,1152]`, domain `aaa` (72 hosts, 8 sub-domains) converts to nodeCount 576 while `bbb` (144 hosts, 2 sub-domains) converts to 144, so the root goes 720 -> 1296 and stops, emitting 72 base blocks where 1296 % 1152 = 144. A sweep of 784 two-domain shapes over that list hits it in 22, and the output then declares `BlockSizes=18,144,1152` over a block list that does not tile into 1152. It needs unequal sub-domain counts per domain, so `tests/models/dual-level.yaml` does not catch it on any block-size list I tried - but that is the shape a real OCI cluster with uneven racks per fabric domain produces. (pkg/translate/block_tree.go:299)
2. `buildNFDObjects` skips both `xclr-domain` and `xclr-sub-domain` when `nvidia.com/gpu.clique` is present, so a provider-supplied sub-domain attribute is not published. This PR's own `TestBuildNFDObjectsSuppressesSubDomainWhenGPUCliqueExists` asserts that suppression, and `docs/reference/node-labels.md` describes it correctly, so this line is the outlier. (docs/engines/nfd.md:152)
3. This location string is the Go field name, while the other two are operator-facing config keys (`fabricLabels[N]` and `acceleratorLabel`). Pointing `acceleratorLabel` at the sub-domain key yields `configured for both acceleratorLabel and XclrSubDomain`, which names a parameter operators cannot set. (pkg/engines/k8s/labeler.go:61)
4. The comment says each depth receives the smallest `blockSize >= max actualNodeCount`, but `getDesiredCountByLevel` calls `pow2GroupCapacity`, which returns the smallest `2^n x base`. With `blockSizes=[18,144]` and a depth max of 20 that is 36, not 144, so the documented rule and the code disagree on a value that drives block padding. The same claim is repeated on `setDesiredCountByLevel`. (pkg/topology/domain.go:140)
5. This entry records what simulation models moved away from, and that key was `network.topology.nvidia.com/accelerator` - `xclr.topology.nvidia.com/domain` did not exist when that change landed. The rename swept up a historical reference, so the entry now describes models leaving a key they never used. (CHANGELOG.md:32)
6. `asBlockVertex` is sound only while every `*Vertex` in `Vertices` is `&child.Vertex`, but `Vertices` is an exported field on an exported type, so nothing outside this file enforces that. A `children map[string]*BlockVertex` field, or an unexported child map with `ChildAt` as the only reader, would let the compiler hold the invariant instead of a doc comment. (pkg/topology/domain.go:60)

pkg/translate/block_tree.go
ravisoundar: Fixed it now, and also added a test case to validate it.

docs/engines/nfd.md
ravisoundar: Fixed it

pkg/engines/k8s/labeler.go
ravisoundar: Fixed now.

pkg/topology/domain.go
ravisoundar: Fixed now.

CHANGELOG.md
ravisoundar: Fixed

pkg/topology/domain.go
ravisoundar: Fixed

pkg/translate/block_tree.go
```diff
@@ -213,159 +224,256 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
-		case *aggregateBlockNode:
-			count += c.nodeCount
-		case *baseBlockNode:
-			count += c.nodeCount
+		sortHostsByName(childHosts)
+
+		// Flush pending hosts into a base block once their count exceeds half the base block size.
+		if len(pendingHosts) > baseBlockSize/2 {
+			blockName := strings.Join(pendingNames, "+")
+			bb := newBaseBlock(blockName, pendingHosts, baseBlockSize)
```
ArangoGutierrez: Strategy 2 packs hosts from several sub-domains into one base block and names it with a `+`-joined compound, but `formatBlockName` errors when nodes in one block derive different names. Both features are new in this release, so a dual-level cluster that sets `blockName.nodeNameRegexp`/`format` fails topology generation outright. Reproduced with one domain, four racks of two hosts, `blockSizes=[8,16]`, and format `$rack`: block001 came out as `r1+r2+r3` and `formatBlockNames` returned `nodes in block "block001" (domain "r1+r2+r3") produce different block names "r1" and "r2"`. Either skip the combining when a block-name formatter is configured, or make the formatter tolerate a combined block.

pkg/translate/block_tree.go
```diff
@@ -213,159 +224,256 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
-		result[i] = child
-		switch c := child.(type) {
-		case *aggregateBlockNode:
-			count += c.nodeCount
-		case *baseBlockNode:
-			count += c.nodeCount
+		sortHostsByName(childHosts)
+
+		// Flush pending hosts into a base block once their count exceeds half the base block size.
+		if len(pendingHosts) > baseBlockSize/2 {
```
ArangoGutierrez: The flush fires once the already-accumulated count passes `baseBlockSize/2` rather than looking ahead at the next child, so a block closes early. The docstring above says it flushes "whenever the running host count reaches baseBlockSize", and the SDD says "No host slots are wasted" — neither matches. Concretely, four sub-domains of two hosts with `baseBlockSize=8` produce two blocks (6 hosts, then 2) where one full block would do. Worth either fixing the threshold or correcting both descriptions.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 - Helm `kubeClient.qps` and `kubeClient.burst` values for tuning the DRA provider and the Kubernetes, NFD, and Slinky engine clients through deployment-level `KUBE_QPS` and `KUBE_BURST` settings.
+- The Kubernetes engine now publishes `xclr.topology.nvidia.com/sub-domain` when a provider supplies `InstanceTopology.XclrSubDomainID`.
+- The NFD engine now publishes separate `xclr-domain` and `xclr-sub-domain` attributes and groups.
+- The graph engine now includes `xclr.topology.nvidia.com/sub-domain` in instance labels when supplied alongside an XCLR domain.
+- Exported Go constant `topology.KeyTopologyXclrSubDomain` for the `xclr.topology.nvidia.com/sub-domain` label key.
+- DRA provider and Slinky engine `kubeQPS` and `kubeBurst` parameters for tuning their independent Kubernetes client rate limits on large clusters.
```
ArangoGutierrez: This entry announces `kubeQPS` and `kubeBurst` parameters on the DRA provider and Slinky engine, but no such parameters exist — the only matches in the tree are the `KUBE_QPS`/`KUBE_BURST` env vars in `internal/k8s/client.go` and the Helm `kubeClient.qps`/`kubeClient.burst` values. The entry five lines above already describes the real mechanism correctly, so this one looks like it should just be dropped.

pkg/topology/domain.go
```diff
@@ -64,3 +87,81 @@ func (m DomainMap) AddHostInfo(hostInfo *HostInfo) {
...
+			placed := 0
+			for _, host := range hosts {
+				gn := host.SubDomain
+				if gn == "" {
+					// A host with no SubDomain in a domain where other hosts carry one
+					// indicates a partially-configured provider (e.g. OCI missing rack
+					// info). Place it in a fallback sub-domain vertex keyed by the
+					// accelerator domain name so the host is always emitted rather than
+					// silently dropped.
+					klog.Warningf("domain %q: host %q has no SubDomain; placing in fallback sub-domain %q", domain, host.HostName, domain)
```
ArangoGutierrez: This warning sits inside the per-host loop, so a partially-racked domain logs one line per node on every regeneration — and the node observer retriggers generation on node and pod changes. The SDD names partially-racked OCI domains as an expected condition, so this is normal operation rather than an anomaly. Consider aggregating to one line per domain with a count.

pkg/providers/oci/provider_sim_test.go
```diff
@@ -254,8 +254,8 @@ func TestSimClientGetComputeHostRackAdditionalData(t *testing.T) {
 
 			instanceTopology, err := convertComputeHost(&resp.ComputeHost)
 			require.NoError(t, err)
-			require.Equal(t, tc.parentDomain, instanceTopology.ParentAcceleratorID)
-			require.Equal(t, tc.parentDomain+"."+tc.rack, instanceTopology.AcceleratorID)
+			require.Equal(t, tc.domain, instanceTopology.XclrDomainID)
+			require.Equal(t, tc.domain+"."+tc.rack, instanceTopology.XclrSubDomainID)
```
ArangoGutierrez, line 258: Both cases here carry a rack (`rack01`, `rack12`), so the rackless branch of `convertComputeHost` is never exercised. Deleting the `if rack != ""` guard in `instance_topology.go` leaves `go test ./pkg/providers/oci/` green — I checked. That guard implements documented behavior: `docs/providers/oci.md` says hosts without rack metadata keep single-level topology, and without it every rackless host gets the sub-domain `<GpuMemoryFabricId>.` with a trailing dot. A third case with an empty rack asserting `XclrSubDomainID` stays empty would cover it.

ArangoGutierrez (commented):
Went through the graph conversion, the labeler upgrade path, and the block packing. The dual-level model itself holds up — I verified the three ToGraph cases and confirmed no host is dropped or duplicated across even, uneven, tiny-sub-domain, and one-level layouts. One blocker: the new sub-domain combining and the new blockName.format feature are mutually incompatible.

1. [must-fix] Strategy 2 packs hosts from several sub-domains into one base block and names it with a `+`-joined compound, but `formatBlockName` errors when nodes in one block derive different names. Both features are new in this release, so a dual-level cluster that sets `blockName.nodeNameRegexp`/`format` fails topology generation outright. Reproduced with one domain, four racks of two hosts, `blockSizes=[8,16]`, and format `$rack`: block001 came out as `r1+r2+r3` and `formatBlockNames` returned `nodes in block "block001" (domain "r1+r2+r3") produce different block names "r1" and "r2"`. Either skip the combining when a block-name formatter is configured, or make the formatter tolerate a combined block. (pkg/translate/block_tree.go:411)
2. [should-fix] The flush fires once the already-accumulated count passes `baseBlockSize/2` rather than looking ahead at the next child, so a block closes early. The docstring above says it flushes "whenever the running host count reaches baseBlockSize", and the SDD says "No host slots are wasted" — neither matches. Concretely, four sub-domains of two hosts with `baseBlockSize=8` produce two blocks (6 hosts, then 2) where one full block would do. Worth either fixing the threshold or correcting both descriptions. (pkg/translate/block_tree.go:409)
3. [should-fix] This entry announces `kubeQPS` and `kubeBurst` parameters on the DRA provider and Slinky engine, but no such parameters exist — the only matches in the tree are the `KUBE_QPS`/`KUBE_BURST` env vars in `internal/k8s/client.go` and the Helm `kubeClient.qps`/`kubeClient.burst` values. The entry five lines above already describes the real mechanism correctly, so this one looks like it should just be dropped. (CHANGELOG.md:16)
4. [should-fix] This warning sits inside the per-host loop, so a partially-racked domain logs one line per node on every regeneration — and the node observer retriggers generation on node and pod changes. The SDD names partially-racked OCI domains as an expected condition, so this is normal operation rather than an anomaly. Consider aggregating to one line per domain with a count. (pkg/topology/domain.go:140)
5. [should-fix] Both cases here carry a rack (`rack01`, `rack12`), so the rackless branch of `convertComputeHost` is never exercised. Deleting the `if rack != ""` guard in `instance_topology.go` leaves `go test ./pkg/providers/oci/` green — I checked. That guard implements documented behavior: `docs/providers/oci.md` says hosts without rack metadata keep single-level topology, and without it every rackless host gets the sub-domain `<GpuMemoryFabricId>.` with a trailing dot. A third case with an empty rack asserting `XclrSubDomainID` stays empty would cover it. (pkg/providers/oci/provider_sim_test.go:258)
6. [consider] This BREAKING entry is a verbatim duplicate of the one five lines above, so the simulation-model annotation change is listed twice in the same Changed section. (CHANGELOG.md:35)
7. [consider] A few comments still describe an earlier block-tree design. This one says buildBlockTree "assigns DesiredNodeCount to every node via a BFS pass" — there is no `DesiredNodeCount` field anywhere and `GetDomainTree` computes `ActualNodeCount`/`MaxChildNodeCount` inline, which its own docstring says avoids a separate BFS pass. Same category: the strategy-3 docstring says the recursion passes `child.MaxChildNodeCount()` when `recurseChildrenIntoAggregate` passes the parent's value, and the `placed` counter is described as excluding unplaced hosts when the fallback branch places every host, so it always equals `len(hosts)`. (pkg/translate/block_complement.go:22; also pkg/translate/block_tree.go:272, pkg/topology/domain.go:152)
8. [consider] The first argument to `validate` is a human-readable config location — the other call sites pass `fabricLabels[0]` and `acceleratorLabel` — but this one passes the label key itself. Setting `fabricLabels: ["xclr.topology.nvidia.com/sub-domain"]` yields: `topology label key "xclr.topology.nvidia.com/sub-domain" is configured for both fabricLabels[0] and xclr.topology.nvidia.com/sub-domain`. The collision is still caught; only the message is confusing. (pkg/engines/k8s/labeler.go:61)

pkg/translate/block_tree.go
ravisoundar: Now the code is updated to skip combining into a single block when the `formatBlockName` is set.

pkg/translate/block_tree.go
ravisoundar: Documentation updated and the code is updated to look ahead to the full block size before flushing the block.

CHANGELOG.md
ravisoundar: Fixed now.

pkg/topology/domain.go
ravisoundar: The warning is moved out of the loop and modified to use a single log statement per domain.

pkg/providers/oci/provider_sim_test.go
ravisoundar: Added a. new test to cover the rackless branch.

pkg/translate/block_tree.go
```diff
@@ -213,159 +224,281 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
+
+	//Wrap all base blocks into an aggregate block node and return.
+	agg := &aggregateBlockNode{id: src.ID}
+	for _, b := range blocks {
+		agg.children = append(agg.children, b)
+		agg.nodeCount += baseBlockSize
 	}
-	return &aggregateBlockNode{children: children, nodeCount: count}
+	return agg
 }
```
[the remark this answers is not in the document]
ravisoundar: In the example above, `maxSiblingNodes = 15` is not correct. When there are 5 domains with 3 nodes each the maxSiblingsNodes will be at the max 3 nodes. Added a new test `TestComplementStrategy2OddSubDomainCount` to validate the above scenario, and it works

pkg/providers/oci/instance_topology.go
```diff
@@ -161,13 +161,10 @@ func convertHost(instanceID, localBlockID, networkBlockID, hpcIslandID, gpuMemor
...
-		domain := *gpuMemoryFabricID
+		topo.XclrDomainID = *gpuMemoryFabricID
 
-		if rack == "" {
-			topo.AcceleratorID = domain
-		} else {
-			topo.AcceleratorID = domain + "." + rack
-			topo.ParentAcceleratorID = domain
+		if rack != "" {
+			topo.XclrSubDomainID = topo.XclrDomainID + "." + rack
```
ArangoGutierrez, line 167: Moving the rack out of the domain ID and into XclrSubDomainID silently coarsens Slurm and Slinky block output when BlockSizes is not configured. Sub-domain data only reaches the block tree through complementBlocks, which returns the blocks unchanged on an empty BlockSizes, so initBlocks groups by accelerator domain alone. Measured on a one-fabric, two-rack cluster with BlockSizes unset: main emits block001=fabric-1.rack-a and block002=fabric-1.rack-b with BlockSizes=2,4; this branch emits a single block001=fabric-1 with BlockSizes=4. Setting BlockSizes=[2,4] restores the two blocks, so the fix is either a fallback to sub-domain granularity on the unconfigured path, or a CHANGELOG breaking entry naming the BlockSizes value OCI operators must now set.

docs/reference/node-labels.md
```diff
@@ -147,7 +155,8 @@ transformers:
     allowedLabels:
       # ... existing labels ...
       # Topograph topology labels (requires Topograph deployed in the cluster)
-      - "network.topology.nvidia.com/accelerator"
+      - "xclr.topology.nvidia.com/domain"
+      - "xclr.topology.nvidia.com/sub-domain"
```
ArangoGutierrez, line 159: The allowedLabels example picks up the two new xclr keys, but the paragraph above it still states that the four network.topology.nvidia.com/* labels ship in NVSentinel's default allowedLabels with no operator configuration required. That no longer holds for accelerator locality after the rename: the xclr keys are not in NVSentinel's defaults, so operators who take the paragraph at its word will quietly lose accelerator metadata on health events.

ArangoGutierrez (commented):
One blocking issue: with BlockSizes unconfigured, OCI block topology loses the rack level that main emits today.

1. [must-fix] Moving the rack out of the domain ID and into XclrSubDomainID silently coarsens Slurm and Slinky block output when BlockSizes is not configured. Sub-domain data only reaches the block tree through complementBlocks, which returns the blocks unchanged on an empty BlockSizes, so initBlocks groups by accelerator domain alone. Measured on a one-fabric, two-rack cluster with BlockSizes unset: main emits block001=fabric-1.rack-a and block002=fabric-1.rack-b with BlockSizes=2,4; this branch emits a single block001=fabric-1 with BlockSizes=4. Setting BlockSizes=[2,4] restores the two blocks, so the fix is either a fallback to sub-domain granularity on the unconfigured path, or a CHANGELOG breaking entry naming the BlockSizes value OCI operators must now set. (pkg/providers/oci/instance_topology.go:167)
2. [should-fix] The allowedLabels example picks up the two new xclr keys, but the paragraph above it still states that the four network.topology.nvidia.com/* labels ship in NVSentinel's default allowedLabels with no operator configuration required. That no longer holds for accelerator locality after the rename: the xclr keys are not in NVSentinel's defaults, so operators who take the paragraph at its word will quietly lose accelerator metadata on health events. (docs/reference/node-labels.md:159)
3. [consider] Validate reports the sub-domain key under the location name xclrSubDomainLabel, but no such engine parameter exists — the key is fixed and the schema says so. An operator who trips the duplicate-key error will go looking through values for a parameter they cannot set. (pkg/engines/k8s/labeler.go:61)
4. [consider] The fallback sub-domain vertex is keyed by the accelerator domain name. If a provider ever emits a SubDomain equal to its Domain, the hosts with no sub-domain merge into that real sub-domain vertex and the warning below misreports where they landed. A key that cannot collide keeps the fallback visible. (pkg/topology/domain.go:144)
5. [consider] The comment points at commit 654a715, which exists only on this branch. The repo squash-merges to main, so that SHA will not resolve for anyone reading this later — naming the release or the PR would age better. (pkg/engines/k8s/kubernetes.go:183)

pkg/translate/block_complement.go
```diff
@@ -59,10 +39,17 @@ func (nt *NetworkTopology) complementBlocks(blocks []*blockInfo, blockSizes []in
 		return blocks
 	}
 
+	if len(blockSizes) < 1 {
```
dmitsh, line 42: **P1: Preserve inferred `BlockSizes` through rendering**

When the configured list is empty, the inferred sizes are assigned only to this local variable. Both output paths later call `getBlockSizes` with the original empty configuration (`block.go:74` and `yaml.go:218`), which derives sizes from the complemented live-node lists and counts an empty placeholder as size zero.

Reproduced with two accelerator domains whose rack populations are `8+8` and `8`: inference selects `[8,16]`, complementing produces block node counts `[8,8,8,0]`, but the rendered footer is `BlockSizes=0,0,0`. Unequal non-empty racks also emit sizes different from those used to construct the tree.

Please compute the effective sizes once and propagate them through both complementing and rendering, or return the inferred sizes alongside the complemented blocks.

pkg/translate/block_tree.go
```diff
@@ -213,159 +224,329 @@ func newEmptyBaseBlock(baseBlockSize int) *baseBlockNode {
...
+	if len(pendingHosts) > 0 {
+		blockName := strings.Join(pendingNames, "+")
+		bb := newBaseBlock(blockName, pendingHosts, baseBlockSize)
+		bb.domain = blockName
+		blocks = append(blocks, bb)
 	}
-	if levels == 1 {
-		return newEmptyBaseBlock(blockSizes[0])
+
+	//Pad the list of blocks to numBaseBlocks with empty base blocks
```
dmitsh: **P1: Pad greedy results to an accelerator-domain aggregate boundary**

`numBaseBlocks` is only enforced as a lower bound here. If indivisible sub-domain groups make the greedy packer produce more blocks, the excess is accepted without rounding the domain slot to a representable aggregate boundary.

For example, with `BlockSizes=[8,32]`, a domain containing five 3-node sub-domains produces three base blocks (`6,6,3` nodes). If the next domain contains four 2-node sub-domains, its real block lands in position 4. Slurm consequently interprets blocks 1-4 as one 32-node aggregate containing nodes from both accelerator domains, defeating the position-based domain grouping this complementing is intended to preserve.

The produced block count needs to be rounded/padded to a valid uniform domain-slot boundary (or the packing strategy must otherwise guarantee it cannot exceed the reserved slot).

pkg/translate/block_complement.go
ravisoundar: done

pkg/translate/block_tree.go
ravisoundar: Removed the greedy logic, and added test `TestComplementOverflowRoundedToAggregateBoundary` to validate the above case.

dmitsh (approved):

PR #417, opened 2026-07-22
feat(demos): add interactive Kubernetes deployment demos

dmitsh:


demos/test-nfd/values.nfd.kwok.yaml
```diff
@@ -0,0 +1,30 @@
...
+  params:
+    modelFileName: medium.yaml
+    generateResponseCode: 202
+    topologyResponseCode: 200
+engine:
+  name: nfd
+  params:
+    nodeSelector:
+      kwok.x-k8s.io/node: "fake"
+    namespace: "node-feature-discovery"
```
ArangoGutierrez: The nfd engine reads its namespace from the NFD_NAMESPACE env var (populated from the top-level nfdNamespace value), not from engine.params.namespace, so this key is silently ignored. It only works here because it matches the default nfdNamespace; changing it would have no effect. Consider dropping it and setting top-level nfdNamespace instead.

demos/dra-slinky/values.dra-slinky.kwok.yaml
```diff
@@ -0,0 +1,40 @@
+provider:
+  name: dra
+  params:
+    nodeSelector:
+      kwok.x-k8s.io/node: "fake"
+    modelFileName: medium.yaml
```
ArangoGutierrez: The dra provider only reads nodeSelector from its params; modelFileName, generateResponseCode, and topologyResponseCode are test-provider keys that config decoding silently drops. modelFileName is especially misleading since dra discovers live topology from node labels rather than a model file, so these three keys are worth removing.

demos/dra-slinky/demo.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
ArangoGutierrez, line 1: This adds demos/ as a new top-level directory. AGENTS.md and .claude/CLAUDE.md ask that a new top-level directory be reflected in the repository map in the same PR, so adding a demos/ row keeps the agent-facing layout in sync.

ArangoGutierrez (commented):
Found 3 items in the new demo tooling — two dead config keys and a docs-sync note. Nice, tidy demo scripts otherwise.

1. The nfd engine reads its namespace from the NFD_NAMESPACE env var (populated from the top-level nfdNamespace value), not from engine.params.namespace, so this key is silently ignored. It only works here because it matches the default nfdNamespace; changing it would have no effect. Consider dropping it and setting top-level nfdNamespace instead. (demos/test-nfd/values.nfd.kwok.yaml:12)
2. The dra provider only reads nodeSelector from its params; modelFileName, generateResponseCode, and topologyResponseCode are test-provider keys that config decoding silently drops. modelFileName is especially misleading since dra discovers live topology from node labels rather than a model file, so these three keys are worth removing. (demos/dra-slinky/values.dra-slinky.kwok.yaml:6)
3. This adds demos/ as a new top-level directory. AGENTS.md and .claude/CLAUDE.md ask that a new top-level directory be reflected in the repository map in the same PR, so adding a demos/ row keeps the agent-facing layout in sync. (demos/dra-slinky/demo.sh:1)

dmitsh:
> Found 3 items in the new demo tooling — two dead config keys and a docs-sync note. Nice, tidy demo scripts otherwise.

thank you, addressed.

ArangoGutierrez (approved):
Thanks for the follow-up round — the substantive items are all addressed. `deploy_topograph` and the NFD chart install are idempotent now (`helm upgrade --install`, `helm repo add --force-update`); `delete_cluster` and the NFD control-plane node name derive from `KUBE_CONTEXT` instead of hardcoded values; every `kubectl` carries `--context`; and the `dra`/`nfd` values files dropped the keys their provider/engine silently ignore. SPDX headers match the repo convention and `demos/` is reflected in AGENTS.md and .claude/CLAUDE.md.

One optional, non-blocking note: in `update-labels.sh` the `kubectl get nodes ... | while read` pipeline runs under `set -e` without `set -o pipefail`, so a failure of the upstream `kubectl` is masked and the script exits 0 having applied no labels. Fine for an interactive, step-gated demo, but adding `pipefail` (as `deploy-slinky.sh` already does) would surface that failure. Approving.

Issue #386, opened 2026-07-09
RBAC: engine/provider-gate the main Topograph ClusterRole (node/pod/daemonset verbs)


ArangoGutierrez:
## Problem

The main api-server `ClusterRole` (`charts/topograph/templates/rbac.yaml`) grants its `nodes`, `pods`, and `daemonsets` verbs **unconditionally** — gated on nothing but `.Values.rbac.create`. Only the `pods/exec` and `configmaps` rules are engine/provider-aware. As a result, a deployment whose selected engine and provider never call the Kubernetes API (any `slurm`/`graph` engine paired with a non-Kubernetes provider, e.g. the default `test` provider) still receives a **cluster-wide `nodes:update`** (a write) plus `pods` and `daemonsets` reads it never exercises. This is the largest least-privilege gap on the primary API surface (audit finding F2).

## Evidence (re-verified against the repo this session, on branch `fix/rbac-drop-unused-verbs`)

Post-PR1 template renders these four rules with no engine/provider gate (`charts/topograph/templates/rbac.yaml:32-51`):

```yaml
rules:
- apiGroups: [""]
  resources: [pods]
  verbs: [list]
- apiGroups: [""]
  resources: [nodes]
  verbs: [get,list,update]
- apiGroups: [apps]
  resources: [daemonsets]
  verbs: [get]
{{- if eq .Values.global.engine.name "slinky" }}
- apiGroups: [""]
  resources: [configmaps]
  verbs: [create,get,update]
{{- end }}
```

Every Kubernetes call the **main api-server binary** (`cmd/topograph`) can make is confined to four loaders — verified by enumerating clientset construction (`grep 'kubernetes.NewForConfig\|InClusterConfig'` over `pkg/ cmd/ internal/`, non-test): the k8s engine (`pkg/engines/k8s/engine.go:60`), the slinky engine (`pkg/engines/slinky/engine.go:124`), the dra provider (`pkg/providers/dra/provider.go:55`), and the infiniband-k8s provider (`pkg/providers/infiniband/provider_k8s.go:57`). Neither `cmd/topograph` nor `pkg/server` builds a client. Per-verb call sites (all confirmed by reading the files):

| Verb (post-PR1) | Actually needed when | Verified call site |
|---|---|---|
[...]

PR #422, opened 2026-07-23
fix(topology): canonicalize compute instance ordering

### LINKED ISSUE #421: [FEA]: canonicalize ComputeInstances discovery and provider fan-out order
### Version

`main` at `7607854c8d6b6509b52a9296d7ce149968c011ba` (2026-07-22).

### Installation method

Source inspection and unit-test analysis.

### Describe the bug

Equivalent sets of compute instances do not currently have a canonical discovery order before provider calls. Several paths derive ordered slices from unordered maps or from upstream list encounter order:

- `pkg/engines/slurm/slurm.go:128-153`: `aggregateComputeInstances` ranges over the `i2n` map and appends regions on first encounter. The returned `[]ComputeInstances` region order therefore depends on Go map iteration. `TestAggregateComputeInstances` uses `ElementsMatch`, which masks this property.
- `internal/k8s/utils.go:107-133`: `GetComputeInstances` preserves the first-seen region order from `NodeList.Items` instead of defining a canonical region order.
- `pkg/engines/slinky/engine.go:249-282`: the Slinky-specific builder repeats the same first-seen-region behavior.
- `pkg/topology/request.go:122-130`: `GetNodeNameList` ranges over each `Instances` map without sorting. This list feeds both bare-metal `pdsh` and `getIbTree`, so the host/probe sequence can vary for the same logical request.
- `pkg/providers/aws/instance_topology.go:61-66`: `DescribeInstanceTopologyInput.InstanceIds` is built directly from an `Instances` map.

The final Slurm serializers already sort many topology structures, so this report does **not** claim that every successful run emits different `topology.conf` bytes. The nondeterminism occurs earlier: region API fan-out order, explicit provider request order, InfiniBand probe-host order, logs/metrics, and which error is returned first when more than one region or probe can fail.

### Minimum reproducible example

A deterministic regression test can construct the same logical mapping through every permutation of input node order, for example:

```go
instances := map[string]string{
    "i-b": "node-b",
    "i-a": "node-a",
    "i-c": "node-c",
}
regions := map[string]string{
    "node-b": "region-b",
    "node-a": "region-a",
    "node-c": "region-b",
}
```

Today, the Slurm aggregation test cannot assert one exact result because region creation follows map iteration. Likewise, a fake AWS client or fake `IBNetDiscover` runner can observe different ID/host call order unless the caller sorts first.

### Expected deterministic behavior

Define one canonical ordering at the `ComputeInstances` boundary:

1. regions sorted lexicographically;
2. instance IDs sorted lexicographically within each region whenever converted to a slice;
3. node names sorted lexicographically for host-oriented command/probe lists.

The ordering should be applied before external provider/API/exec calls, not only during final serialization. Canonicalization must not mutate the caller-owned request and must not change the canonical `topology.Graph` shape.

### Proposed implementation

- Add small shared sorting helpers near `topology.ComputeInstances` rather than duplicating ordering logic across engines/providers.
- Make Slurm, Kubernetes, and Slinky aggregation return regions in the same canonical order.
- Make `GetNodeNameList` return a sorted list while preserving its current membership semantics.
- Sort AWS explicit `InstanceIds` before `DescribeInstanceTopology`; audit other map-to-request conversions in DSX/OCI/InfiniBand and use the same helper where applicable.
- Keep provider discovery sequential unless concurrency is separately designed; deterministic sequencing also makes first-error precedence stable.

### Acceptance criteria

- [ ] Slurm, Kubernetes, and Slinky produce identical, exactly ordered `[]ComputeInstances` for every permutation of the same logical nodes.
- [ ] `GetNodeNameList` has an exact-order test; callers receive the same host sequence for permuted map construction.
- [ ] Fake AWS and InfiniBand clients record the same request/probe sequence for equivalent inputs.
- [ ] Multi-region failure tests establish deterministic first-error precedence.
- [ ] Tests enumerate/permutate inputs and assert exact traces; they do not rely on running a map iteration repeatedly and hoping the runtime changes order.
- [ ] Existing generated topology output remains byte-identical for the same successful logical input.
- [ ] `make qualify` passes, documentation impact is evaluated, and the user-visible determinism fix is recorded under `CHANGELOG.md` `[Unreleased]`.

### Alternatives considered

Sorting only the final `topology.conf` or JSON output is insufficient because provider calls and failure selection have already occurred. Relying on Kubernetes list order is also insufficient because it would leave the Slurm and Go-map paths unordered and would make engine behavior inconsistent.

---

By submitting this issue, I agree to follow the project's Code of Conduct and contributing guidelines. I searched open and closed issues and PRs for ComputeInstances ordering, map iteration, provider call order, `GetNodeNameList`, InfiniBand probe order, and region ordering; I found no duplicate.

teerthsharma:
## Description

Closes #421.

Go map iteration is unspecified, but several discovery paths converted maps to ordered provider inputs without sorting. The same logical `ComputeInstances` set could therefore change:

- region/API fan-out order;
- AWS `InstanceIds` order;
- InfiniBand probe-host order; and
- which error is returned first when multiple regions fail.

This is observable before final serialization in retries, logs, metrics, probes, and error precedence. It does not claim that every successful run produced different `topology.conf` bytes.

## Changes

- Add a shared, non-mutating `CanonicalComputeInstances` total order.
- Apply it at the request, Kubernetes, Slinky, Slurm, and AWS boundaries.
- Sort AWS explicit instance IDs and host-oriented node lists.
- Replace order-insensitive assertions with exact traces across explicit permutations.
- Record the fix in `CHANGELOG.md`.

The canonical `topology.Graph` shape and provider/engine responsibilities are unchanged.

## Validation

Fresh on PR head `969fad9` after merging current `main`:

- `go vet` on all 7 affected packages: pass
- focused `go test` on all 7 affected packages: pass
- `git diff --check`: pass
- exact tests cover K8s, Slinky, Slurm, AWS, InfiniBand, deterministic first-error precedence, and caller-input non-mutation

NVIDIA runner workflows are pending the repository's external-contributor validation step. Security validation found a reliability/reproducibility defect, not a security vulnerability: membership, call count, command content, privileges, and exposed data are unchanged.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - `kwok-nodes` now prevents model-provided metadata from overriding its required KWOK selector and model-derived `topograph.nvidia.com/instance` and `topograph.nvidia.com/region` annotations.
 - Slinky dynamic-node reconciliation now reuses listed Node annotations, skips unchanged nodes without a per-node GET, and patches only changed topology annotations, substantially reducing Kubernetes client-side throttling on large clusters.
 - Corrected DRA provider guidance to document its Slinky-only block-topology scope, dependency on pre-existing `nvidia.com/gpu.clique` labels, and inability to guide placement across NVLink partitions without backend-fabric topology.
+- Compute-instance discovery and provider fan-out now use deterministic region ordering, including AWS API requests and InfiniBand probes.
 - The NFD engine now rejects an empty generated object set when cleanup is enabled, preserving the last published topology instead of deleting every Topograph-managed NFD object after an empty provider result or over-narrow node selection.
 - The NFD engine now groups nodes with `nvidia.com/gpu.clique` by that authoritative accelerator value instead of omitting their accelerator attribute.
 - The NFD engine now publishes the `system.name/nodename` attribute required by NFD to populate `NodeFeatureGroup.status.nodes`, including for simulated KWOK nodes where no NFD worker executes.
--- a/internal/k8s/utils.go
+++ b/internal/k8s/utils.go
@@ -130,5 +130,5 @@ func GetComputeInstances(nodes *corev1.NodeList) []topology.ComputeInstances {
 		cis = append(cis, topology.ComputeInstances{Region: region, Instances: regions[region]})
 	}
 
-	return cis
+	return topology.CanonicalComputeInstances(cis)
 }
--- a/pkg/engines/k8s/kubernetes_test.go
+++ b/pkg/engines/k8s/kubernetes_test.go
@@ -50,7 +50,21 @@ func TestGetComputeInstances(t *testing.T) {
 		},
 		{
 			name:  "Case 3: valid input",
-			nodes: &corev1.NodeList{Items: []corev1.Node{node1, node2, node3}},
+			nodes: &corev1.NodeList{Items: []corev1.Node{node3, node2, node1}},
+			cis: []topology.ComputeInstances{
+				{
+					Region:    "r1",
+					Instances: map[string]string{"i1": "node1", "i2": "node2"},
+				},
+				{
+					Region:    "r2",
+					Instances: map[string]string{"i3": "node3"},
+				},
+			},
+		},
+		{
+			name:  "Case 4: valid input permutation",
+			nodes: &corev1.NodeList{Items: []corev1.Node{node2, node1, node3}},
 			cis: []topology.ComputeInstances{
 				{
 					Region:    "r1",
--- a/pkg/engines/k8s/labeler.go
+++ b/pkg/engines/k8s/labeler.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"maps"
 	"slices"
 	"strings"
 
@@ -102,8 +103,8 @@ func (l *topologyLabeler) ApplyNodeLabels(ctx context.Context, graph *topology.G
 		return err
 	}
 
-	for nodeName, labels := range nodeMap {
-		if err := labeler.AddNodeLabels(ctx, nodeName, labels); err != nil {
+	for _, nodeName := range slices.Sorted(maps.Keys(nodeMap)) {
+		if err := labeler.AddNodeLabels(ctx, nodeName, nodeMap[nodeName]); err != nil {
 			return err
 		}
 	}
--- a/pkg/engines/slinky/engine.go
+++ b/pkg/engines/slinky/engine.go
@@ -298,7 +298,7 @@ func getComputeInstances(nodes *corev1.NodeList, nodeMap map[string]string) ([]t
 		cis = append(cis, topology.ComputeInstances{Region: region, Instances: regions[region]})
 	}
 
-	return cis, nil
+	return topology.CanonicalComputeInstances(cis), nil
 }
 
 func withGPUCliqueDomains(graph *topology.Graph, clusterNodes *clusterNodes) (*topology.Graph, *httperr.Error) {
--- a/pkg/engines/slinky/engine_test.go
+++ b/pkg/engines/slinky/engine_test.go
@@ -334,7 +334,21 @@ func TestGetComputeInstances(t *testing.T) {
 		},
 		{
 			name:  "Case 3: valid input",
-			nodes: &corev1.NodeList{Items: []corev1.Node{node1, node2, node3, nodeNone}},
+			nodes: &corev1.NodeList{Items: []corev1.Node{node3, node2, node1, nodeNone}},
+			cis: []topology.ComputeInstances{
+				{
+					Region:    "r1",
+					Instances: map[string]string{"i1": "node1", "i2": "node2"},
+				},
+				{
+					Region:    "r2",
+					Instances: map[string]string{"i3": "node3"},
+				},
+			},
+		},
+		{
+			name:  "Case 4: valid input permutation",
+			nodes: &corev1.NodeList{Items: []corev1.Node{node2, nodeNone, node1, node3}},
 			cis: []topology.ComputeInstances{
 				{
 					Region:    "r1",
--- a/pkg/engines/slurm/slurm.go
+++ b/pkg/engines/slurm/slurm.go
@@ -152,7 +152,7 @@ func aggregateComputeInstances(i2n, nodeRegions map[string]string) []topology.Co
 	}
 	klog.V(4).Infof("Detected regions: %v", regions)
 
-	return cis
+	return topology.CanonicalComputeInstances(cis)
 }
 
 func GetNodeList(ctx context.Context) ([]string, error) {
--- a/pkg/engines/slurm/slurm_test.go
+++ b/pkg/engines/slurm/slurm_test.go
@@ -79,7 +79,7 @@ func TestAggregateComputeInstances(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			cis := aggregateComputeInstances(tc.i2n, tc.regions)
-			require.ElementsMatch(t, tc.cis, cis)
+			require.Equal(t, tc.cis, cis)
 		})
 	}
 }
--- a/pkg/engines/slurm/slurm_test.go
+++ b/pkg/engines/slurm/slurm_test.go
@@ -79,7 +79,7 @@ func TestAggregateComputeInstances(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			cis := aggregateComputeInstances(tc.i2n, tc.regions)
-			require.Equal(t, tc.cis, cis)
+			require.ElementsMatch(t, tc.cis, cis)
 		})
 	}
 }
--- a/pkg/providers/aws/instance_topology.go
+++ b/pkg/providers/aws/instance_topology.go
@@ -19,7 +19,9 @@ package aws
 import (
 	"context"
 	"fmt"
+	"maps"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
@@ -35,7 +37,7 @@ var defaultPageSize int32 = 100
 func (p *baseProvider) generateInstanceTopology(ctx context.Context, pageSize *int, cis []topology.ComputeInstances) (*topology.ClusterTopology, *httperr.Error) {
 	topo := topology.NewClusterTopology()
 
-	for _, ci := range cis {
+	for _, ci := range topology.CanonicalComputeInstances(cis) {
 		if err := p.generateRegionInstanceTopology(ctx, pageSize, &ci, topo); err != nil {
 			return nil, err
 		}
@@ -60,10 +62,7 @@ func (p *baseProvider) generateRegionInstanceTopology(ctx context.Context, pageS
 	// AWS allows up to 100 explicitly specified instance IDs
 	if n := len(ci.Instances); n <= 100 {
 		klog.Infof("Getting instance topology for %d instances", n)
-		input.InstanceIds = make([]string, 0, n)
-		for instanceID := range ci.Instances {
-			input.InstanceIds = append(input.InstanceIds, instanceID)
-		}
+		input.InstanceIds = slices.Sorted(maps.Keys(ci.Instances))
 	} else {
 		klog.Infof("Getting instance topology with page size %d", pageSize)
 		input.MaxResults = client.PageSize()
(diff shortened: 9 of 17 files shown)


ArangoGutierrez (commented):
Thanks for the thorough writeup. The audit in #421 is accurate: every path you list really does derive an ordered slice from an unordered source, and the line references check out. The permutation cases plus the non-mutation assertions in `pkg/server/engine_test.go` and `pkg/topology/request_test.go` are the right technique for this class of property; asserting against an exact recorded trace, rather than re-running a map loop and hoping the runtime reorders, is how you actually test it. You also scoped the claim carefully in the PR body instead of asserting that successful runs emit different bytes, which is the version that would not have survived.

I traced each of the five paths to its terminus before writing this.

**Paths 1, 2, 3** (slurm `aggregateComputeInstances`, `internal/k8s.GetComputeInstances`, slinky `getComputeInstances`) all feed `prv.GenerateTopologyConfig`, and for every `ClusterTopology`-based provider that ends in `ToGraph` at `pkg/topology/graph.go:87`. `ToGraph` inserts each vertex into maps keyed by ID: `forest` at `graph.go:93`, `nodes[tierKey]` at `graph.go:98` and `graph.go:134-142`, `treeRoot.Vertices` via `maps.Copy` at `graph.go:169-172`. No append position survives that.

**Path 4, pdsh half.** `provider_bm.go:36` hands the list to `exec.Pdsh`, which passes it through `cluset.Compact` at `internal/exec/exec.go:63`. `Compact` sorts prefixes at `internal/cluset/cluset.go:54` and numeric suffixes at `cluset.go:60-62`. The `-w` argument is canonical before it reaches pdsh, so the `sort.Strings` you added at `request.go:159` is erased one call downstream.

**Path 4, IB half.** Probe order is genuinely observable: `getIbTree` at `pkg/providers/infiniband/common.go:32` contacts the first unvisited host. But the result lands in `rootMap`, read back in map order at `common.go:58`, and `NewMerger` re-sorts by vertex ID at `pkg/topology/merger.go:33-35`. A failed probe is `klog.Warningf` and `continue` at `common.go:35-37`, never an error.

**Path 5 (AWS `InstanceIds`).** This one does change bytes on the wire. The response is then selected by map lookup at `instance_topology.go:87` and appended to `ClusterTopology.Instances`, terminating in the same `ToGraph`. Worth noting AWS passes `normalize=false` at `pkg/providers/aws/provider.go:191`, so the sort in `Normalize` (`graph.go:218-236`) never runs on this path, and the output is stable regardless.

I checked that empirically rather than by reading alone. On `main` at `edb20aa`, without this PR, permuting both the region fan-out order and the within-region append order produces a deep-equal `*topology.Graph` and byte-identical `topology.conf` under `topology/tree` and `topology/block`, at both `normalize=false` and `normalize=true`. Since `GenerateOutput` takes only the graph and its params (`pkg/engines/engines.go:31`), an identical `Graph` means identical output for slurm, k8s, nfd, and slinky alike.

That leaves the three claims that survive the byte-identical objection, which are the interesting ones. First-error precedence is real: `generateInstanceTopology` returns on the first failing region at `instance_topology.go:38-41`, so with two broken regions the message names an arbitrary one. The IB probe order and the AWS request order are as described above.

So the question I would put back to you, for each of those three: what does an operator see that is different, and what would they do differently as a result? For first-error precedence, the API returns 502 either way, and `processRequestWithRetries` at `pkg/server/engine.go:54-76` retries the whole request (502 is retryable at `internal/httpreq/httpreq.go:44`) up to `maxRetries = 5` at `engine.go:36`. There is no per-region retry whose behavior depends on which region failed first. For the other two I could not find the symptom at all. I also went looking for the historical cost, since a property like this usually leaves a trail: the only ordering work in this repo is #207 (`fix(slurm): fix block order`) and #13, both fixed in `pkg/translate` at the serializer, neither traceable to discovery order. No flaky-test issue, no support issue, no retry logic keyed on order.

On cost, in maintainer terms. The rule here is that added complexity has to buy something a user or operator can observe, and this lands a comparator in `pkg/topology`, the one package we deliberately do not change casually because every provider and engine depends on its shape. Two specifics. The tie-break half of `compareComputeInstances` (`request.go:65-76`, everything after the `Region` compare) is unreachable through every in-tree producer: slurm keys by region at `slurm.go:132`, k8s at `internal/k8s/utils.go:108` with `122-123`, slinky at `slinky/engine.go:270` with `290-291`. None can emit two entries sharing a `Region`, so that code is exercised only by its own unit test. And converting `ElementsMatch` to `Equal` at `slurm_test.go:82` and `request_test.go:185` prices in every future ordering change; those assertions were not masking a bug, they were encoding that order is not part of the contract.

There is also a consistency gap worth naming: the change does not fully close the property it is named after. The k8s engine's outbound writes still go out in map order at `pkg/engines/k8s/labeler.go:105`, and `pkg/providers/dsx/instance_topology.go:26-30` and `pkg/providers/oci/provider_imds.go:71-74` still build request arguments straight from map iteration. Closing those would roughly double the surface, and by the argument above it would not buy an observable symptom either. (Greptile's note about the second `CanonicalComputeInstances` at `instance_topology.go:40` is correct as well; the slice is already canonical from `pkg/server/engine.go:146`.)

Where that leaves me: I think the right call for the project is to close #422 and relabel #421 as a feature request we keep on file. If a real reproducibility requirement shows up later (a byte-stable audit trail of outbound API calls, or a replay harness), the audit you wrote is exactly the document we would start from, and it will still be accurate then. That is worth more parked and correct than merged and unexercised. You reached the "feature request, not bug" read yourself in #421; this is the same conclusion applied to the code.

If you want something here with clear value, #386 is open and unassigned and would suit you: engine/provider-gate the main Topograph `ClusterRole`. Today the api-server holds cluster-wide `nodes:update` plus `pods` and `daemonsets` reads regardless of the configured engine and provider, including combinations that make zero Kubernetes API calls. The issue already carries the per-verb call-site table, an acceptance-criteria list, and the one real design decision (omit the `ClusterRole` and its binding entirely for zero-rule combos, rather than rendering empty rules), so there is nothing ambiguous left to negotiate. It stacks on #384, which is merged, so you have a worked example of the same gating pattern in the same template. Scope is `charts/topograph/templates/rbac.yaml` plus helm-unittest cases, and the benefit is one an operator can see in `kubectl auth can-i`. I will review it promptly.

Say the word and it is yours. Either way, send the second PR.

teerthsharma:
> Thanks for the thorough writeup. The audit in #421 is accurate: every path you list really does derive an ordered slice from an unordered source, and the line references check out. The permutation cases plus the non-mutation assertions in `pkg/server/engine_test.go` and `pkg/topology/request_test.go` are the right technique for this class of property; asserting against an exact recorded trace, rather than re-running a map loop and hoping the runtime reorders, is how you actually test it. You also scoped the claim carefully in the PR body instead of asserting that successful runs emit different bytes, which is the version that would not have survived.
> 
> I traced each of the five paths to its terminus before writing this.
> 
> **Paths 1, 2, 3** (slurm `aggregateComputeInstances`, `internal/k8s.GetComputeInstances`, slinky `getComputeInstances`) all feed `prv.GenerateTopologyConfig`, and for every `ClusterTopology`-based provider that ends in `ToGraph` at `pkg/topology/graph.go:87`. `ToGraph` inserts each vertex into maps keyed by ID: `forest` at `graph.go:93`, `nodes[tierKey]` at `graph.go:98` and `graph.go:134-142`, `treeRoot.Vertices` via `maps.Copy` at `graph.go:169-172`. No append position survives that.
> 
> **Path 4, pdsh half.** `provider_bm.go:36` hands the list to `exec.Pdsh`, which passes it through `cluset.Compact` at `internal/exec/exec.go:63`. `Compact` sorts prefixes at `internal/cluset/cluset.go:54` and numeric suffixes at `cluset.go:60-62`. The `-w` argument is canonical before it reaches pdsh, so the `sort.Strings` you added at `request.go:159` is erased one call downstream.
> 
> **Path 4, IB half.** Probe order is genuinely observable: `getIbTree` at `pkg/providers/infiniband/common.go:32` contacts the first unvisited host. But the result lands in `rootMap`, read back in map order at `common.go:58`, and `NewMerger` re-sorts by vertex ID at `pkg/topology/merger.go:33-35`. A failed probe is `klog.Warningf` and `continue` at `common.go:35-37`, never an error.
> 
> **Path 5 (AWS `InstanceIds`).** This one does change bytes on the wire. The response is then selected by map lookup at `instance_topology.go:87` and appended to `ClusterTopology.Instances`, terminating in the same `ToGraph`. Worth noting AWS passes `normalize=false` at `pkg/providers/aws/provider.go:191`, so the sort in `Normalize` (`graph.go:218-236`) never runs on this path, and the output is stable regardless.
> 
> I checked that empirically rather than by reading alone. On `main` at `edb20aa`, without this PR, permuting both the region fan-out order and the within-region append order produces a deep-equal `*topology.Graph` and byte-identical `topology.conf` under `topology/tree` and `topology/block`, at both `normalize=false` and `normalize=true`. Since `GenerateOutput` takes only the graph and its params (`pkg/engines/engines.go:31`), an identical `Graph` means identical output for slurm, k8s, nfd, and slinky alike.
> 
> That leaves the three claims that survive the byte-identical objection, which are the interesting ones. First-error precedence is real: `generateInstanceTopology` returns on the first failing region at `instance_topology.go:38-41`, so with two broken regions the message names an arbitrary one. The IB probe order and the AWS request order are as described above.
> 
> So the question I would put back to you, for each of those three: what does an operator see that is different, and what would they do differently as a result? For first-error precedence, the API returns 502 either way, and `processRequestWithRetries` at `pkg/server/engine.go:54-76` retries the whole request (502 is retryable at `internal/httpreq/httpreq.go:44`) up to `maxRetries = 5` at `engine.go:36`. There is no per-region retry whose behavior depends on which region failed first. For the other two I could not find the symptom at all. I also went looking for the historical cost, since a property like this usually leaves a trail: the only ordering work in this repo is #207 (`fix(slurm): fix block order`) and #13, both fixed in `pkg/translate` at the serializer, neither traceable to discovery order. No flaky-test issue, no support issue, no retry logic keyed on order.
> 
> On cost, in maintainer terms. The rule here is that added complexity has to buy something a user or operator can observe, and this lands a comparator in `pkg/topology`, the one package we deliberately do not change casually because every provider and engine depends on its shape. Two specifics. The tie-break half of `compareComputeInstances` (`request.go:65-76`, everything after the `Region` compare) is unreachable through every in-tree producer: slurm keys by region at `slurm.go:132`, k8s at `internal/k8s/utils.go:108` with `122-123`, slinky at `slinky/engine.go:270` with `290-291`. None can emit two entries sharing a `Region`, so that code is exercised only by its own unit test. And converting `ElementsMatch` to `Equal` at `slurm_test.go:82` and `request_test.go:185` prices in every future ordering change; those assertions were not masking a bug, they were encoding that order is not part of the contract.
> 
> There is also a consistency gap worth naming: the change does not fully close the property it is named after. The k8s engine's outbound writes still go out in map order at `pkg/engines/k8s/labeler.go:105`, and `pkg/providers/dsx/instance_topology.go:26-30` and `pkg/providers/oci/provider_imds.go:71-74` still build request arguments straight from map iteration. Closing those would roughly double the surface, and by the argument above it would not buy an observable symptom either. (Greptile's note about the second `CanonicalComputeInstances` at `instance_topology.go:40` is correct as well; the slice is already canonical from `pkg/server/engine.go:146`.)
> 
> Where that leaves me: I think the right call for the project is to close #422 and relabel #421 as a feature request we keep on file. If a real reproducibility requirement shows up later (a byte-stable audit trail of outbound API calls, or a replay harness), the audit you wrote is exactly the document we would start from, and it will still be accurate then. That is worth more parked and correct than merged and unexercised. You reached the "feature request, not bug" read yourself in #421; this is the same conclusion applied to the code.
> 
> If you want something here with clear value, #386 is open and unassigned and would suit you: engine/provider-gate the main Topograph `ClusterRole`. Today the api-server holds cluster-wide `nodes:update` plus `pods` and `daemonsets` reads regardless of the configured engine and provider, including combinations that make zero Kubernetes API calls. The issue already carries the per-verb call-site table, an acceptance-criteria list, and the one real design decision (omit the `ClusterRole` and its binding entirely for zero-rule combos, rather than rendering empty rules), so there is nothing ambiguous left to negotiate. It stacks on #384, which is merged, so you have a worked example of the same gating pattern in the same template. Scope is `charts/topograph/templates/rbac.yaml` plus helm-unittest cases, and the benefit is one an operator can see in `kubectl auth can-i`. I will review it promptly.
> 
> Say the word and it is yours. Either way, send the second PR.
> 

I will be working as told ,I have fixed almost every caveat of this PR , hoping this will provide good reference for the feature request .I have also started working on the issue told thanks for your time

dmitsh:
Thanks for the contribution. Let's hold off on this PR for now. While the change is useful, it touches quite a few files without addressing a specific functional issue. Given the scope of the changes, we’d prefer to avoid taking this in at this point. We can revisit it later if there’s a stronger need for this feature.

PR #425, opened 2026-07-23
feat(slurm): derive block names from node names

### LINKED ISSUE #424: [FEA] Prefix Slurm block names from node names
### Feature Request

Add a `blockNamePrefixRegexp` option to the Slurm and Slinky engines for `topology/block` output.

The option should match a prefix at the beginning of each node name and prepend that prefix to the generated block name. This allows block names to follow cluster-specific naming conventions and avoids generic names such as `block001`.

### Example

Given these nodes:

```text
srv1101
srv1102
...
srv1108
```

And this engine configuration:

```yaml
engine:
  name: slinky
  params:
    plugin: topology/block
    blockSizes: [18]
    blockNamePrefixRegexp: "^srv[0-9]{2}"
```

The generated block should be named:

```text
BlockName=srv11 Nodes=srv[1101-1108]
```

For per-partition Slinky topology YAML, the equivalent name would be `srv11block1`.

### Requirements

- Support `blockNamePrefixRegexp` in both the Slurm and Slinky engines.
- Support it for:
  - Cluster-wide `topology/block` configuration.
  - Individual entries under `topologies`.
- Use Go regular-expression syntax.
- Only accept matches beginning at the start of the node name.
- Prepend the matched value to the existing deterministic block ID.
- Preserve default block names for unmatched nodes and empty complemented blocks.
- Ensure node topology specifications reference the prefixed block names.
- Reject invalid regular expressions with an actionable configuration error.
- Preserve existing behavior when the option is omitted.
- Document the option and add unit/integration coverage.

dmitsh:
## Description

Add configurable block-name derivation to the Slurm and Slinky engines for `topology/block` output.

Sites can define a regular expression for node names and a format string for the resulting block name:

```
blockName:
  nodeNameRegexp: 'd([0-9]{2})-r([0-9]{2})'
  format: 'domain${1}_rack${2}'
```
For a node such as gpu-d05-r04-srv4, this produces:
```
BlockName=domain05_rack04
```
The option supports both cluster-wide and per-partition topology configurations in the Slurm and Slinky engines. When blockName is configured, both nodeNameRegexp and format are required.

Topology generation returns an error when:
- A node in a non-empty block does not match the expression.
- Nodes in the same block produce different names.
- The format produces an empty name.
- Two different blocks produce the same name.

Empty complemented blocks have no node name to evaluate and retain their generated default names. Dynamic-node topology specifications use the resulting derived block names.

Closes #424


pkg/translate/topology_test.go
```diff
@@ -265,6 +265,34 @@ func TestToBlockTopology(t *testing.T) {
 	require.Equal(t, testBlockConfig1_1, buf.String())
 }
 
+func TestToBlockTopologyWithNodePrefix(t *testing.T) {
```
ArangoGutierrez: All the added prefix tests use `^`-anchored patterns, so `FindStringIndex` always returns a match at index 0 and the `loc[0] == 0` start-of-name guard in `blockNameWithNodePrefix` is never actually exercised: dropping that condition would leave every test green. A case with an unanchored pattern that matches mid-name (e.g. `srv[0-9]{2}` against `xsrv12node`) would lock in that mid-string matches don't get prefixed.

ArangoGutierrez (commented):
One test-coverage note on the new block-name-prefix feature; the code path itself reads correct and is well covered otherwise.

1. All the added prefix tests use `^`-anchored patterns, so `FindStringIndex` always returns a match at index 0 and the `loc[0] == 0` start-of-name guard in `blockNameWithNodePrefix` is never actually exercised: dropping that condition would leave every test green. A case with an unanchored pattern that matches mid-name (e.g. `srv[0-9]{2}` against `xsrv12node`) would lock in that mid-string matches don't get prefixed. (pkg/translate/topology_test.go:268)

dmitsh:
> One test-coverage note on the new block-name-prefix feature; the code path itself reads correct and is well covered otherwise.
Thank you! Resolved.

pkg/translate/block.go
```diff
@@ -55,7 +55,9 @@ func (nt *NetworkTopology) toBlockTopology(wr io.Writer, skeletonOnly bool) *htt
 	// Refresh nodeInfo.blockID so GetNodeTopologySpec returns IDs that match the
 	// emitted topology file. complementBlocks may renumber blocks when it splits
 	// a domain across multiple base blocks, invalidating the IDs set by initBlocks.
+	blockNamePrefixRegexp := compileBlockNamePrefixRegexp(nt.config.BlockNamePrefixRegexp)
 	for _, b := range blocks {
+		b.id = blockNameWithNodePrefix(b.id, b.nodes, blockNamePrefixRegexp)
```
ArangoGutierrez: blockNameWithNodePrefix has no uniqueness guard, so when complementBlocks splits one accelerator domain across multiple base blocks (any domain larger than the base block size), the split siblings share a node prefix and collapse to the same name: topology.conf gets duplicate BlockName= lines, which Slurm's block plugin rejects, and nodeInfo.blockID is overwritten so GetNodeTopologySpec returns the same id for nodes in different blocks. With the documented ^srv[0-9]{2} on a 16-node srv11xx domain at base size 8 this emits two BlockName=srv11 entries (verified against a 12-node domain at base size 4, which produced three identical BlockName=Node1 lines). The per-partition YAML path in getBlockTopologyUnit has the same root cause. Consider disambiguating collided names or rejecting a regexp whose derived names are not unique.

ArangoGutierrez (commented):
Found one issue worth a look; the rest of the change is clean and well-covered.

1. blockNameWithNodePrefix has no uniqueness guard, so when complementBlocks splits one accelerator domain across multiple base blocks (any domain larger than the base block size), the split siblings share a node prefix and collapse to the same name: topology.conf gets duplicate BlockName= lines, which Slurm's block plugin rejects, and nodeInfo.blockID is overwritten so GetNodeTopologySpec returns the same id for nodes in different blocks. With the documented ^srv[0-9]{2} on a 16-node srv11xx domain at base size 8 this emits two BlockName=srv11 entries (verified against a 12-node domain at base size 4, which produced three identical BlockName=Node1 lines). The per-partition YAML path in getBlockTopologyUnit has the same root cause. Consider disambiguating collided names or rejecting a regexp whose derived names are not unique. (pkg/translate/block.go:60)

pkg/translate/block.go
ArmedGuy: My counter-question is actually why does complementBlocks split blocks. If the provider has specified that the nodes are in the same topology, why does Topograph not honor that? If anything, that behavior should be configurable, not default.

ArmedGuy:
I would not lock this on being node prefixes. Different HPC clusters will have different naming setups of their nodes, some describing more complex topologies.

If you want this to be truly flexible for each site, you should specify one Regex with capture groups, and one format string for the output.

As an example:
```

gpu-d05-r04-srv4

regex = "d([0-9]{2}-r([0-9]{2})"
format = "domain$1_rack$2"
```

would yield BlockName=domain05_rack04 (and ideally require BlockSizes=x,y).
A simpler one that is applicable today is:
```
regex = "domain({0-9}[3}"
format = "block$1"
```

so a node named domain001 would be in
`BlockName=block001`

Each site can decide their own format for their needs.

pkg/translate/block.go
dmitsh: > My counter-question is actually why does complementBlocks split blocks. If the provider has specified that the nodes are in the same topology, why does Topograph not honor that? If anything, that behavior should be configurable, not default.

complementBlocks doesn’t split provider-discovered domains by default. It only does that when the user explicitly sets blockSizes. In that case, Topograph assumes the user wants blocks based on blockSizes[0], so it splits larger domains and adds empty blocks when needed to complete the hierarchy.
If blockSizes isn’t set, Topograph keeps the domains discovered by the provider and only calculates the BlockSizes value to output.
The bigger issue is that Slurm’s topology formats are fairly limited. The tree format describes a switch hierarchy, and the block format describes a block hierarchy. Neither can fully represent the actual topology.
Topograph has to fit the discovered topology into one of those formats. A proper long-term solution should come from Slurm rather than adding more workarounds in Topograph.

ArangoGutierrez (commented):
Second pass — the earlier round's items look resolved: block names now enforce cross-block uniqueness, and the regexp/format design reads correct and is well covered. Nothing blocking.

One small optional test suggestion: in `formatBlockName` (`pkg/translate/topology.go`), the empty-name rejection (`blockName format ... produces an empty name for node`) is the only error branch without a test — the unmatched-node, inconsistent-names, and duplicate-name branches each have one. A one-line case with a format that expands to an empty string for a matching node would lock the behavior in.

(Note: this review's summary was briefly posted with unrelated text by mistake and has been corrected.)

ArangoGutierrez (approved):
Redesign looks good — regexp + format covers the flexible per-site naming from the earlier discussion, cross-block uniqueness and the empty-name/mismatch/unmatched paths are all guarded and now tested, and CI is green. LGTM.

PR #427, opened 2026-07-24
fix(slinky): reduce Kubernetes API throttling

### LINKED ISSUE #423: Kubernetes API throttling with the DRA provider in Slinky clusters
### Description
When Topograph runs in a Slinky cluster with the DRA provider, it generates a noticeable volume of Kubernetes API requests. This can trigger API throttling and delay request execution.
Investigate the request pattern, identify the component or code path responsible for the excessive API traffic, and address the root cause.
### Expected outcome
- Measure and characterize the Kubernetes API requests generated in this configuration.
- Identify the root cause of the excessive request volume.
- Reduce unnecessary or duplicate API calls.
- Confirm that throttling no longer materially affects request execution timing.
- Add relevant tests and/or metrics to prevent regressions.

dmitsh:
## Description

Reduce Kubernetes API client-side throttling when running the Slinky engine
with dynamic nodes, particularly on large DRA-backed clusters.

Previously, every reconciliation performed an additional Node GET for each
participating node before updating its topology annotation. This caused API
request volume to scale with cluster size even when annotations were already
current.

This change:

- Reuses Node annotations from the existing cluster-wide LIST.
- Skips unchanged nodes without additional API calls.
- Uses a targeted PATCH for changed topology annotations.
- Adds Slinky Node PATCH RBAC permissions.
- Adds configurable `kubeQPS` and `kubeBurst` parameters for the Slinky engine and DRA provider.
- Rejects negative rate-limit values.
- Adds tests covering API action counts, unchanged reconciliation, PATCH failures, and Helm RBAC.
- Updates Helm values, API documentation, provider/engine documentation, snapshots, and the changelog.

closes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- DRA provider and Slinky engine `kubeQPS` and `kubeBurst` parameters for tuning their independent Kubernetes client rate limits on large clusters.
 - Slurm and Slinky block topology configurations can set `blockName.nodeNameRegexp` and `blockName.format` to derive unique block names from site-specific node naming conventions.
 - Simulation model files may omit empty leaf-switch definitions; the model loader now creates leaf switches referenced by the parent switch hierarchy.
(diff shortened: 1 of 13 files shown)


ArangoGutierrez (commented):
No issues found — checked for bugs and CLAUDE.md compliance.

PR #434, opened 2026-07-27
ci(k8s): add kind integration tests for simulated providers

### LINKED ISSUE #433: [TEST] Add K8s and Slinky end-to-end tests to PR checks
### Description
Topograph’s current PR checks cover Go unit tests and Helm chart rendering, but they do not verify a complete K8s and Slinky deployment.

Regressions can therefore pass CI even when they affect interactions between the Helm chart, API server, node observer, provider translation, K8s node labeling, and Slinky configmaps.

Add an end-to-end PR check using kind and KWOK that:
- Builds the Topograph image from the PR.
- Creates virtual nodes from an existing simulation model.
- Deploys Topograph through the Helm chart.
- Exercises the K8s and Slinky engines in simulation.
- Verifies the expected results.
- Prints useful cluster diagnostics when a test fails.

The test should use pinned dependency versions and require no cloud credentials or physical infrastructure.

dmitsh:
## Description
Add a Kubernetes integration workflow for PR checks using kind and KWOK.
The workflow:
- Builds and loads the current Topograph image into kind.
- Creates simulated Kubernetes nodes from model files.
- Tests the K8s engine with the test, aws-sim, and gcp-sim providers.
- Tests the Slinky engine with the dra, and oci-sim providers.
- Verifies the expected outcome.
- Collects pod logs, events, and node labels on failure.
- Pins kind and KWOK versions for reproducible runs.
closes #433


faganihajizada (dismissed):
Looks good!

PR #438, opened 2026-07-28
fix(k8s): reduce client-side throttling and centralize Kubernetes client rate limits

### LINKED ISSUE #437: Kubernetes API throttling with the K8S and NFD providers
Large Kubernetes clusters can experience significant client-side throttling during topology reconciliation.

The Kubernetes engine currently performs a per-node GET followed by a full Node update. With client-go defaults of 5 QPS and burst 10, reconciling thousands of nodes can take several minutes.
Kubernetes client rate limits are also configured independently across the DRA provider and Kubernetes-backed engines, requiring duplicated Helm values. The NFD engine does not currently accept these settings and constructs typed and dynamic clients with separate rate limiters.

dmitsh:
## Description
Large-cluster topology reconciliation could spend several minutes waiting on client-go’s default rate limiter. The Kubernetes engine performed a per-node GET followed by a full Node UPDATE, generating roughly two API requests per node even when topology labels were already current.

This PR reduces Kubernetes API traffic and adds consistent rate-limit configuration across Kubernetes-backed components.

### Changes

- Rework Kubernetes engine label reconciliation to:
  - Reuse the Node list fetched during topology discovery.
  - Skip nodes whose topology labels are already current.
  - PATCH only changed labels instead of performing a GET and full UPDATE.
  - Continue removing stale Topograph-managed labels.
  - Preserve existing `nvidia.com/gpu.clique` behavior.
- Change Kubernetes engine RBAC from Node `update` to `patch`.
- Add `kubeQPS` and `kubeBurst` support to the Kubernetes and NFD engines.
- Configure NFD’s typed and dynamic Kubernetes clients with one shared token-bucket limiter, making QPS an aggregate limit across both clients.
- Centralize client rate-limit normalization for the DRA provider and Kubernetes, NFD, and Slinky engines.
  - When only QPS or burst is configured, the other value uses the client-go default.
- Add shared Helm configuration:
  ```yaml
  kubeClient:
    qps: 50
    burst: 100
  ```
- Update Helm values, schema, NOTES, examples, API documentation, engine/provider documentation, and the changelog.

closes #437


charts/topograph/tests/configmap_test.yaml
```diff
@@ -51,3 +51,44 @@ tests:
...
+      provider:
+        name: dra
+      engine:
+        name: k8s
+      kubeClient:
+        qps: 50
+    asserts:
+      - matchRegex:
+          path: data["topograph-config.yaml"]
+          pattern: "(?s)providerParams:.*kubeQPS: 50.*kubeBurst: 10.*engineParams:.*kubeQPS: 50.*kubeBurst: 10"
```
ArangoGutierrez: `kubeBurst: 10` is an unanchored substring that also matches `kubeBurst: 100`, so this case cannot pin the client-go burst default it exists to prove. Changing the template to `default 100` renders `kubeBurst: 100` here and all 120 chart tests still pass; anchoring the value would catch it.

docs/api.md
```diff
@@ -101,11 +99,13 @@ Topograph exposes three endpoints for interacting with the service. Below are th
...
       - **useDynamicNodes**: (optional) Used in: [`slinky`]. If `true`, Kubernetes nodes matched by the Node Selector will be annotated with the topology spec.
       - **useGpuCliqueLabel**: (optional) Used in: [`slinky`]. If `true`, `topology/block` domains are built from the GPU Operator's `nvidia.com/gpu.clique` node label instead of provider accelerator-domain data.
-      - **kubeQPS**: (optional) Used in: [`slinky`]. A positive number overriding the engine Kubernetes client's default request rate of 5 QPS.
-      - **kubeBurst**: (optional) Used in: [`slinky`]. A positive integer overriding the engine Kubernetes client's default burst capacity of 10.
       - **configUpdateMode**: (optional) Used in: [`slinky`]. By default, the full topology YAML is written in the Slurm ConfigMap. `skeleton-only` overrides to include switches or blocks only (no node lines); `none` skips updating the topology key in the ConfigMap.
   - **nodes**: (optional) Supplies the cluster nodes used for topology generation as an array of regions mapping instance IDs to node names.
 
+For Helm deployments, configure Kubernetes client rate limits only through
```
ArangoGutierrez: This PR adds `kubeQPS` and `kubeBurst` as request parameters on the k8s and NFD engines, but the API reference loses its four existing entries rather than gaining two more. They stay live request parameters - the server applies config `providerParams`/`engineParams` only when the payload omits the key - so direct API callers have nothing documenting them.

internal/k8s/client_test.go
```diff
@@ -0,0 +1,64 @@
...
+			require.Equal(t, tc.wantQPS, config.QPS)
+			require.Equal(t, tc.wantBurst, config.Burst)
+			require.Nil(t, config.RateLimiter)
+		})
+	}
+}
+
+func TestConfigureSharedClientRateLimiter(t *testing.T) {
+	config := &rest.Config{}
+	ConfigureSharedClientRateLimiter(config, 0, 0)
```
ArangoGutierrez: This only calls the function with `(0, 0)`, so nothing checks that a configured qps/burst actually reaches the shared token bucket. Rewriting the body to `normalizedClientRateLimits(0, 0)`, which discards both arguments and would make `kubeClient.qps` a no-op for NFD, leaves this package and `pkg/engines/nfd` green.

ArangoGutierrez (commented):
Found 6 issues. The rate-limit centralization and the PATCH-only reconciliation both look right; the notes below are mostly about guards that stopped discriminating, plus one default that shifts.

1. `kubeBurst: 10` is an unanchored substring that also matches `kubeBurst: 100`, so this case cannot pin the client-go burst default it exists to prove. Changing the template to `default 100` renders `kubeBurst: 100` here and all 120 chart tests still pass; anchoring the value would catch it. (charts/topograph/tests/configmap_test.yaml:80)
2. This PR adds `kubeQPS` and `kubeBurst` as request parameters on the k8s and NFD engines, but the API reference loses its four existing entries rather than gaining two more. They stay live request parameters - the server applies config `providerParams`/`engineParams` only when the payload omits the key - so direct API callers have nothing documenting them. (docs/api.md:105)
3. This only calls the function with `(0, 0)`, so nothing checks that a configured qps/burst actually reaches the shared token bucket. Rewriting the body to `normalizedClientRateLimits(0, 0)`, which discards both arguments and would make `kubeClient.qps` a no-op for NFD, leaves this package and `pkg/engines/nfd` green. (internal/k8s/client_test.go:60)
4. Flipping this case to a positive assertion for k8s leaves slurm, nfd, and graph uncovered, since the `by default` case above already renders with the chart's default k8s engine. Widening the template guard to `ne .Values.engine.name "slinky"` grants those engines node patch and the suite stays green; a `notContains` case with `engine.name: slurm` would restore it. (charts/topograph/tests/rbac_test.yaml:152)
5. The comment documents that this always installs a bucket, but the NFD loader calls it unconditionally, so an unconfigured deployment now shares one 5 QPS bucket where the typed and dynamic clients each built their own before - roughly 10 QPS aggregate down to 5, in a PR aimed at reducing throttling. An early return on zero/zero would keep the old default, or a CHANGELOG note if the lower aggregate is intended. (internal/k8s/client.go:26)
6. A node in the graph but outside the engine `nodeSelector` was labeled through the old GET path; it now errors, and `ApplyNodeLabels` stops on the first error, so the request fails with a 502 after some nodes are already patched. Logging and skipping would let the pass finish for the nodes in scope. (pkg/engines/k8s/kubernetes.go:43)

charts/topograph/tests/rbac_test.yaml
```diff
@@ -149,19 +149,19 @@ tests:
             resources: [nodes]
             verbs: [patch]
 
-  - it: does not grant node patch access to non-slinky engines
+  - it: grants node patch access to the k8s engine
```
ArangoGutierrez, line 152: Flipping this case to a positive assertion for k8s leaves slurm, nfd, and graph uncovered, since the `by default` case above already renders with the chart's default k8s engine. Widening the template guard to `ne .Values.engine.name "slinky"` grants those engines node patch and the suite stays green; a `notContains` case with `engine.name: slurm` would restore it.

internal/k8s/client.go
```diff
@@ -0,0 +1,41 @@
...
+		return
+	}
+	qps, burst = normalizedClientRateLimits(qps, burst)
+	config.QPS = qps
+	config.Burst = burst
+}
+
+// ConfigureSharedClientRateLimiter always installs one token bucket, using
+// client-go defaults for zero values, so clients created from config share it.
+func ConfigureSharedClientRateLimiter(config *rest.Config, qps float32, burst int) {
```
ArangoGutierrez: The comment documents that this always installs a bucket, but the NFD loader calls it unconditionally, so an unconfigured deployment now shares one 5 QPS bucket where the typed and dynamic clients each built their own before - roughly 10 QPS aggregate down to 5, in a PR aimed at reducing throttling. An early return on zero/zero would keep the old default, or a CHANGELOG note if the lower aggregate is intended.

pkg/engines/k8s/kubernetes.go
```diff
@@ -26,31 +29,98 @@ func (eng *K8sEngine) GetComputeInstances(ctx context.Context, _ any) ([]topolog
...
 func (eng *K8sEngine) AddNodeLabels(ctx context.Context, nodeName string, labels map[string]string) error {
-	klog.Infof("Applying labels on node %s : %v", nodeName, labels)
-	node, err := eng.client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err := eng.loadNodes(ctx); err != nil {
+		return err
+	}
+
+	node, ok := eng.cachedNodeMap[nodeName]
+	if !ok {
+		return fmt.Errorf("node %q was not found in the selected Kubernetes nodes", nodeName)
```
ArangoGutierrez: A node in the graph but outside the engine `nodeSelector` was labeled through the old GET path; it now errors, and `ApplyNodeLabels` stops on the first error, so the request fails with a 502 after some nodes are already patched. Logging and skipping would let the pass finish for the nodes in scope.

docs/api.md
dmitsh: We changed QPS and Burst to be the deployment-level settings, instead of repeating them in engines/providers.
There are no provider or engine request parameters for QPS/Burst.

pkg/node_observer/status_informer_test.go
```diff
@@ -236,6 +243,39 @@ func TestDeduplicatesRequests(t *testing.T) {
...
+	done := make(chan error, 1)
+
+	go func() {
+		done <- s.Start()
+	}()
+
+	select {
+	case err := <-done:
+		require.Failf(t, "Start returned before Stop", "error: %v", err)
+	case <-time.After(50 * time.Millisecond):
```
ArangoGutierrez, line 344: This 50 ms window is shorter than client-go's 100 ms WaitForCacheSync poll period, so `Start` has not returned yet either way and the case passes unchanged against the old `go s.run()`. Mutating the implementation back to `go s.run()` keeps this green; widening the window to 500 ms fails that mutant and still passes the real one.

pkg/engines/k8s/kubernetes.go
```diff
@@ -26,31 +29,102 @@ func (eng *K8sEngine) GetComputeInstances(ctx context.Context, _ any) ([]topolog
...
-	klog.Infof("Applying labels on node %s : %v", nodeName, labels)
-	node, err := eng.client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err := eng.loadNodes(ctx); err != nil {
+		return err
+	}
+
+	node, ok := eng.cachedNodeMap[nodeName]
+	if !ok {
+		if len(eng.params.NodeSelector) != 0 {
+			klog.Warningf("Skipping topology labels for node %q because it does not match the engine nodeSelector", nodeName)
```
ArangoGutierrez: With the per-node GET gone, this branch cannot tell a node outside the nodeSelector from one that does not exist, so a wrong node name in a request `nodes` payload is swallowed and `GenerateOutput` still answers OK. Counting the skipped nodes and surfacing the total would keep a genuine typo visible.

docs/api.md
ArangoGutierrez: `kubeQPS` and `kubeBurst` are still live request parameters: the mapstructure tags remain on the DRA provider and the k8s, NFD and Slinky engines, `getParameters` validates them with a 400, and config params apply only when the payload omits the key, so a payload value wins over `kubeClient`. This PR adds two more of them and drops the last entries documenting any, which the doc-impact table asks to keep in sync.

pkg/node_observer/status_informer.go
```diff
@@ -132,13 +133,14 @@ func (s *StatusInformer) Start() error {
 		return err
 	}
 
-	go s.run()
-
+	s.run()
```
ArangoGutierrez, line 151: `Start` now blocks on `run`, but `run` only re-checks `stopCh` between `process` calls and the new `s.cancel()` reaches only the informer factories. A SIGTERM landing mid-retry waits out `DoRequestWithRetries`, whose `time.Sleep` and timeout-less `http.Client` observe the root context that `Stop` never cancels, so shutdown stalls up to the 5 minute `maxRetryAfter`.

ArangoGutierrez (commented):
Second pass over the changes since e5f17d0. The five points from the first round all landed, and I mutation-checked four of them: anchoring `kubeBurst: 10$` now catches a `default 100` template change, the shared-limiter test goes red when the function discards its arguments, and widening the RBAC guard to `ne "slinky"` now fails two cases. Five things left, one of them a test that does not discriminate.

1. This 50 ms window is shorter than client-go's 100 ms WaitForCacheSync poll period, so `Start` has not returned yet either way and the case passes unchanged against the old `go s.run()`. Mutating the implementation back to `go s.run()` keeps this green; widening the window to 500 ms fails that mutant and still passes the real one. (pkg/node_observer/status_informer_test.go:267)
2. With the per-node GET gone, this branch cannot tell a node outside the nodeSelector from one that does not exist, so a wrong node name in a request `nodes` payload is swallowed and `GenerateOutput` still answers OK. Counting the skipped nodes and surfacing the total would keep a genuine typo visible. (pkg/engines/k8s/kubernetes.go:44)
3. `kubeQPS` and `kubeBurst` are still live request parameters: the mapstructure tags remain on the DRA provider and the k8s, NFD and Slinky engines, `getParameters` validates them with a 400, and config params apply only when the payload omits the key, so a payload value wins over `kubeClient`. This PR adds two more of them and drops the last entries documenting any, which the doc-impact table asks to keep in sync. (docs/api.md:105)
4. `Start` now blocks on `run`, but `run` only re-checks `stopCh` between `process` calls and the new `s.cancel()` reaches only the informer factories. A SIGTERM landing mid-retry waits out `DoRequestWithRetries`, whose `time.Sleep` and timeout-less `http.Client` observe the root context that `Stop` never cancels, so shutdown stalls up to the 5 minute `maxRetryAfter`. (pkg/node_observer/status_informer.go:136)
5. The only test for `requestOnAPIServerDelete` passes a live `*corev1.Pod`, so the new `DeletedFinalStateUnknown` branch never runs: deleting that whole case leaves the package green. Relists after a watch gap deliver tombstones, which is the rolling-deployment path this handler was added for. (pkg/node_observer/status_informer_test.go:169)

pkg/node_observer/status_informer_test.go
```diff
@@ -165,6 +166,12 @@ func TestAPIServerPodReadinessRequiresTargetContainer(t *testing.T) {
 	))
 }
 
+func TestAPIServerPodDeleteQueuesRequest(t *testing.T) {
```
ArangoGutierrez, line 170: The only test for `requestOnAPIServerDelete` passes a live `*corev1.Pod`, so the new `DeletedFinalStateUnknown` branch never runs: deleting that whole case leaves the package green. Relists after a watch gap deliver tombstones, which is the rolling-deployment path this handler was added for.

pkg/engines/k8s/kubernetes.go
```diff
@@ -7,50 +7,146 @@ package k8s
...
 func (eng *K8sEngine) AddNodeLabels(ctx context.Context, nodeName string, labels map[string]string) error {
-	klog.Infof("Applying labels on node %s : %v", nodeName, labels)
-	node, err := eng.client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err := eng.loadNodes(ctx); err != nil {
+		return err
+	}
+
+	node, ok := eng.cachedNodeMap[nodeName]
+	if !ok {
+		return fmt.Errorf("node %q was not found in Kubernetes", nodeName)
```
ArangoGutierrez: A cache miss returns a hard error, and ApplyNodeLabels returns on the first error, so one unresolvable node aborts the whole labeling pass and the request 502s. When no nodes are supplied in the request the cache holds only the nodeSelector-filtered Nodes, but the graph can legitimately contain others: infiniband/provider_k8s.go builds Domains from its own GetNodes(p.params.nodeListOpt) and ignores the compute instances passed in. The previous per-node GET found those Nodes cluster-wide and labeled them. Consider warn-and-skip here, matching the nodeSelector branch just below.

pkg/engines/k8s/kubernetes.go
```diff
@@ -7,50 +7,146 @@ package k8s
...
 )
 
-func (eng *K8sEngine) GetComputeInstances(ctx context.Context, _ any) ([]topology.ComputeInstances, *httperr.Error) {
-	nodes, err := k8s.GetNodes(ctx, eng.client, eng.params.nodeListOpt)
+func (eng *K8sEngine) ResolveComputeInstances(ctx context.Context, instances []topology.ComputeInstances, _ any) ([]topology.ComputeInstances, *httperr.Error) {
+	listOptions := eng.params.nodeListOpt
+	if len(instances) != 0 {
+		// Cache every Node so output generation can distinguish a nonexistent
+		// requested node from one intentionally excluded by nodeSelector.
+		listOptions = nil
```
ArangoGutierrez: Dropping nodeListOpt makes every request that supplies nodes LIST all Nodes in the cluster, and internal/k8s.GetNodes issues a single List with no Limit or Continue. On a large cluster with a narrow selector that is a full unpaged Node list held for the request, which pulls against the throttling reduction this PR is after. Keeping the selector and issuing one targeted Nodes().Get on a cache miss gives the same absent-vs-filtered distinction at bounded cost.

pkg/node_observer/status_informer.go
```diff
@@ -215,6 +231,7 @@ func (s *StatusInformer) startAPIServerInformer() error {
 					s.sendRequest()
 				}
 			},
+			DeleteFunc: s.requestOnAPIServerDelete,
```
ArangoGutierrez, line 234: Nothing asserts this handler is registered. Deleting this line keeps the whole pkg/node_observer suite green, TestAPIServerPodDeleteQueuesRequest included, because that test calls requestOnAPIServerDelete directly on a bare StatusInformer. The sibling fix in the same commit is covered properly, TestStartBlocksUntilStopped fails when s.run() goes back to go s.run(), so this is an isolated gap on the PR's headline changelog fix.

charts/topograph/templates/NOTES.txt
```diff
@@ -47,6 +47,13 @@
   NOTE: The NFD engine writes NodeFeature and NodeFeatureGroup resources in
         {{ .Values.nfdNamespace }}. This must be the namespace where NFD master runs.
 {{- end }}
+{{- $kubeClient := default dict .Values.kubeClient }}
+{{- if or (hasKey $kubeClient "qps") (hasKey $kubeClient "burst") }}
+
+  NOTE: Kubernetes client limits are qps={{ default 5 (get $kubeClient "qps") }}
+        and burst={{ default 10 (get $kubeClient "burst") }}.
+        Configure these limits only through kubeClient.
```
ArangoGutierrez, line 55: These are not the only surface. provider.params.kubeQPS and engine.params.kubeQPS still reach the server through the node-observer request payload, and http_server.go only applies the chart-generated providerParams/engineParams when the payload omits the key, so a request-level value wins. docs/engines/k8s.md:163 states that precedence correctly and docs/providers/dra.md:75 repeats the same only claim. Either soften the wording or add a fail guard in _validation.tpl like the existing env.NFD_NAMESPACE one.

ArangoGutierrez (commented):
Found 4 issues.

1. A cache miss returns a hard error, and ApplyNodeLabels returns on the first error, so one unresolvable node aborts the whole labeling pass and the request 502s. When no nodes are supplied in the request the cache holds only the nodeSelector-filtered Nodes, but the graph can legitimately contain others: infiniband/provider_k8s.go builds Domains from its own GetNodes(p.params.nodeListOpt) and ignores the compute instances passed in. The previous per-node GET found those Nodes cluster-wide and labeled them. Consider warn-and-skip here, matching the nodeSelector branch just below. (pkg/engines/k8s/kubernetes.go:54)
2. Dropping nodeListOpt makes every request that supplies nodes LIST all Nodes in the cluster, and internal/k8s.GetNodes issues a single List with no Limit or Continue. On a large cluster with a narrow selector that is a full unpaged Node list held for the request, which pulls against the throttling reduction this PR is after. Keeping the selector and issuing one targeted Nodes().Get on a cache miss gives the same absent-vs-filtered distinction at bounded cost. (pkg/engines/k8s/kubernetes.go:33)
3. Nothing asserts this handler is registered. Deleting this line keeps the whole pkg/node_observer suite green, TestAPIServerPodDeleteQueuesRequest included, because that test calls requestOnAPIServerDelete directly on a bare StatusInformer. The sibling fix in the same commit is covered properly, TestStartBlocksUntilStopped fails when s.run() goes back to go s.run(), so this is an isolated gap on the PR's headline changelog fix. (pkg/node_observer/status_informer.go:234)
4. These are not the only surface. provider.params.kubeQPS and engine.params.kubeQPS still reach the server through the node-observer request payload, and http_server.go only applies the chart-generated providerParams/engineParams when the payload omits the key, so a request-level value wins. docs/engines/k8s.md:163 states that precedence correctly and docs/providers/dra.md:75 repeats the same only claim. Either soften the wording or add a fail guard in _validation.tpl like the existing env.NFD_NAMESPACE one. (charts/topograph/templates/NOTES.txt:55)

charts/topograph/tests/deployment_test.yaml
```diff
@@ -248,6 +248,42 @@ tests:
...
+      kubeClient:
+        qps: 50
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: KUBE_QPS
+            value: "50"
+        template: templates/deployment.yaml
+      - notContains:
```
ArangoGutierrez, line 281: This `notContains` compares whole array elements, so `{name: KUBE_BURST}` never equals the rendered `{name: KUBE_BURST, value: ...}` and the assertion cannot fail: deleting the `{{- if $hasKubeBurst }}` guard from the template still leaves this suite at 24 passed / 24 total. Adding `any: true` under `content` is the only variant that goes red on that mutant, and it stays green on the current template. The guard is worth covering, because an emitted `KUBE_BURST: ""` makes `envInt` fail and every engine and provider Loader return 502. Same shape as the `rbac_test.yaml` feedback on #402; the two new RBAC `notContains` assertions here are fine as-is, since their content is a complete rule object.

pkg/engines/slinky/engine.go
```diff
@@ -133,11 +129,8 @@ func Loader(_ context.Context, params engines.Config) (engines.Engine, *httperr.
 		return nil, httperr.NewError(http.StatusBadGateway, err.Error())
 	}
 
-	if p.KubeQPS > 0 {
-		config.QPS = p.KubeQPS
-	}
-	if p.KubeBurst > 0 {
-		config.Burst = p.KubeBurst
+	if err := k8s.ConfigureClientRateLimits(config); err != nil {
+		return nil, httperr.NewError(http.StatusBadGateway, err.Error())
```
ArangoGutierrez: Mapping a rate-limit config error to 502 makes it retryable, since `ShouldRetry` includes 502. Measured against `processRequestWithRetries`: a 502 costs 5 attempts and 30s of backoff (2+4+8+16), while the 400 that the removed `kubeQPS`/`kubeBurst` validation returned costs 1 attempt. A permanent misconfiguration fits 400 better. Same mapping in `pkg/engines/k8s/engine.go`, `pkg/engines/nfd/engine.go`, and `pkg/providers/dra/provider.go`.

internal/k8s/client.go
```diff
@@ -0,0 +1,78 @@
...
+
+	qps, burst = normalizedClientRateLimits(qps, burst)
+	config.QPS = qps
+	config.Burst = burst
+	config.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(qps, burst)
+	return nil
+}
+
+func envFloat32(name string) (float32, bool, error) {
+	value, ok := os.LookupEnv(name)
```
ArangoGutierrez, line 48: `LookupEnv` reports a set-but-empty `KUBE_QPS` or `KUBE_BURST` as present, so parsing fails and the loader hard-errors instead of falling back to the default: `KUBE_QPS=""` returns `KUBE_QPS must be a non-negative number`, and a whitespace-only value behaves the same. The chart's schema blocks this, but `docs/engines/k8s.md` tells non-Helm deployments to set these variables directly, where empty is the usual way to express unset. `getNFDNamespace` in the NFD engine already trims and treats blank as absent.

ArangoGutierrez (commented):
Three findings, all in the new rate-limit plumbing. The core of the change looks right: with both variables unset the config is left untouched so client-go's 5 QPS / burst 10 apply, an explicit zero maps to the same 5/10, the k8s engine's label sync is now a strategic-merge Patch so the `update` -> `patch` RBAC tightening is consistent, and the slinky annotation path was already on Patch.

1. This `notContains` compares whole array elements, so `{name: KUBE_BURST}` never equals the rendered `{name: KUBE_BURST, value: ...}` and the assertion cannot fail: deleting the `{{- if $hasKubeBurst }}` guard from the template still leaves this suite at 24 passed / 24 total. Adding `any: true` under `content` is the only variant that goes red on that mutant, and it stays green on the current template. The guard is worth covering, because an emitted `KUBE_BURST: ""` makes `envInt` fail and every engine and provider Loader return 502. Same shape as the `rbac_test.yaml` feedback on #402; the two new RBAC `notContains` assertions here are fine as-is, since their content is a complete rule object. (charts/topograph/tests/deployment_test.yaml:281)
2. Mapping a rate-limit config error to 502 makes it retryable, since `ShouldRetry` includes 502. Measured against `processRequestWithRetries`: a 502 costs 5 attempts and 30s of backoff (2+4+8+16), while the 400 that the removed `kubeQPS`/`kubeBurst` validation returned costs 1 attempt. A permanent misconfiguration fits 400 better. Same mapping in `pkg/engines/k8s/engine.go`, `pkg/engines/nfd/engine.go`, and `pkg/providers/dra/provider.go`. (pkg/engines/slinky/engine.go:133)
3. `LookupEnv` reports a set-but-empty `KUBE_QPS` or `KUBE_BURST` as present, so parsing fails and the loader hard-errors instead of falling back to the default: `KUBE_QPS=""` returns `KUBE_QPS must be a non-negative number`, and a whitespace-only value behaves the same. The chart's schema blocks this, but `docs/engines/k8s.md` tells non-Helm deployments to set these variables directly, where empty is the usual way to express unset. `getNFDNamespace` in the NFD engine already trims and treats blank as absent. (internal/k8s/client.go:47)

ArangoGutierrez (approved):
All three findings from the previous review are resolved. I checked each fix by mutation rather than by reading the diff:

- `any: true` on the KUBE_BURST assertion: deleting the `{{- if $hasKubeBurst }}` guard now turns the suite red (1 failed, 23 passed), where it previously stayed at 24 passed / 24 total.
- The four Loaders returning 400: making 400 retryable turns Case 3 of `TestProcessRequestWithRetries` red, so the new attempt-count assertions are load-bearing.
- `TrimSpace` plus the blank check: removing the blank check fails all four cases of `TestConfigureClientRateLimitsTreatsBlankEnvironmentAsUnset`. The whitespace-padded numeric cases are a good addition beyond what was raised.

Chart suite is green at 122 tests and 145 snapshots, and the affected Go packages pass. The `deploy-slinky.sh` KWOK wait is guarded against a zero node count, and the demo's `trigger.podSelector` resolves to a real config field with the node-observer RBAC already granting pods list/watch.

One non-blocking note for a follow-up: reverting all four Loaders from 400 back to 502 still leaves the suite green, so the status code choice itself has no coverage. Returning `*httperr.Error` from `ConfigureClientRateLimits` would make it assertable in `internal/k8s` without a cluster, and would keep the decision in one place instead of four. Not worth holding this PR for.

ArangoGutierrez (approved):
All three comments from the last round check out.

- `any: true` repairs the `notContains` assertion. Flipping the template guard to always emit `KUBE_BURST` now fails exactly that one case out of 24, where before it could not fail at all.
- The rate-limit config error is non-retryable now, and at all four loader call sites rather than just Slinky. `ShouldRetry` covers 408, 429, 500, 502, 503 and 504, so 400 is excluded, and `engine_test.go` pins the attempt count at 1.
- Blank and whitespace-only `KUBE_QPS` and `KUBE_BURST` fall back to the defaults. Reverting the guard fails the new blank-environment cases, so it is a real guard.

Go suite and the chart suite are green locally.

One note, not blocking: the new `kubectl wait` in `scripts/deploy-slinky.sh` sits in the `kwok` branch, and CI runs `--compute-mode none`, so that path is not exercised.

PR #439, opened 2026-07-29
Feature/node data broker validating admission policy

### LINKED ISSUE #386: RBAC: engine/provider-gate the main Topograph ClusterRole (node/pod/daemonset verbs)

Shankar-v27:
## Description

Closes #386

This PR updates the Topograph Helm chart to follow the principle of least privilege by rendering only the RBAC permissions required for the configured provider and engine.

### Summary

- Dynamically compute the required `Node` verbs based on the enabled providers and engines.
- Render Kubernetes RBAC rules (`pods`, `pods/exec`, `daemonsets`, `configmaps`, etc.) only when they are required.
- Skip rendering the `ClusterRole` and `ClusterRoleBinding` when no Kubernetes API permissions are needed.
- Update the Helm RBAC unit tests to cover the new permission matrix.
- Update the changelog.

### Motivation

Previously, the chart could grant permissions that were unnecessary for a given deployment configuration. This change minimizes the RBAC footprint by ensuring only the permissions required for the selected configuration are rendered, reducing the attack surface while preserving existing functionality.

## Testing

- Updated Helm RBAC unit tests to validate the generated permission matrix.
- Verified the rendered templates through static review.
- Rebased onto the latest `main` and resolved merge conflicts.

## Checklist

- [x] I have read the project's contributing guidelines.
- [x] Tests have been updated where applicable.
- [x] Documentation has been updated where applicable.
- [x] All commits are signed off (`git commit -s`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Security
 
 - Main Topograph API server ClusterRole rules are gated by the selected engine and provider: `nodes`, `pods`, `daemonsets`, and `configmaps` permissions render only when the engine or provider reaches the Kubernetes API, and the ClusterRole and ClusterRoleBinding are omitted entirely for non-Kubernetes combinations such as the `test` provider with the `slurm` engine.
+- Node RBAC permissions in the main Topograph API server ClusterRole are collapsed into a single rule following least privilege: the Kubernetes engine receives Node `[get, list, patch]`, the dynamic Slinky engine (`useDynamicNodes: true`) receives Node `[list, patch]`, and other node-consuming configurations receive Node `[list]` without unnecessary `patch`, `get`, or `update` access.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Security
 
+- Gated Topograph API server RBAC ClusterRole rules dynamically to render permissions only when required by the configured provider or engine, collapsed node rules into a single rule with dynamically computed verbs, and omitted ClusterRole and ClusterRoleBinding entirely when no Kubernetes permissions are required ([#386](https://github.com/NVIDIA/topograph/issues/386)).
 - Opt-in `ValidatingAdmissionPolicy` and binding to restrict the `node-data-broker` ServiceAccount to only updating its host Node resource, mitigating privilege escalation via node spec or annotation manipulation (audit F1 compliance).
 - Removed unused RBAC verbs from the Topograph API server and node-data-broker ClusterRoles (least-privilege): API server `pods` rule dropped `get` (list-only), `daemonsets` rule dropped `list` (get-only), and the Slinky `configmaps` rule dropped `list`; node-data-broker `nodes` rule dropped `list` (get/update), and the InfiniBand `daemonsets`/`pods` rules dropped `list`/`get` respectively (get-only, list-only).
 - Upgraded `golang.org/x/text` from `v0.38.0` to `v0.39.0` to address a dependency vulnerability.
--- a/charts/topograph/templates/rbac.yaml
+++ b/charts/topograph/templates/rbac.yaml
@@ -65,14 +65,16 @@ rules:
   verbs: [create]
 {{- end }}
 {{- if $needsNodes }}
-- apiGroups: [""]
-  resources: [nodes]
-  verbs: [get,list]
-{{- if or (eq $engine "k8s") (eq $engine "slinky") }}
-- apiGroups: [""]
-  resources: [nodes]
-  verbs: [patch]
+{{- $nodeVerbs := list "list" -}}
+{{- if eq $engine "k8s" -}}
+{{- $nodeVerbs = list "get" "list" "patch" -}}
+{{- end -}}
+{{- if and (eq $engine "slinky") (get (default dict .Values.engine.params) "useDynamicNodes") -}}
+{{- $nodeVerbs = list "list" "patch" -}}
 {{- end }}
+- apiGroups: [""]
+  resources: [nodes]
+  verbs: {{ $nodeVerbs | toJson }}
 {{- end }}
 {{- if $needsDaemonSets }}
 - apiGroups: [apps]
--- a/charts/topograph/templates/rbac.yaml
+++ b/charts/topograph/templates/rbac.yaml
@@ -25,40 +25,49 @@ already supplies nodes, a pod selector, or the default flat topology.
 {{- end -}}
 {{- end -}}
 {{- end -}}
+{{- $needsPods := or (eq .Values.provider.name "infiniband-k8s") (eq .Values.engine.name "slinky") -}}
+{{- $needsDaemonsets := eq .Values.provider.name "infiniband-k8s" -}}
+{{- $needsNodesList := or (eq .Values.provider.name "infiniband-k8s") (eq .Values.provider.name "dra") (eq .Values.engine.name "k8s") (eq .Values.engine.name "slinky") (eq .Values.engine.name "nfd") -}}
+{{- $needsConfigMaps := eq .Values.engine.name "slinky" -}}
+{{- $needsClusterRole := or $needsPods $needsPodExec.value $needsNodesList $needsDaemonsets $needsConfigMaps -}}
+{{- if $needsClusterRole }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ include "topograph.rbacName" . }}
 rules:
+{{- if $needsPods }}
 - apiGroups: [""]
   resources: [pods]
   verbs: [list]
+{{ end }}
 {{- if $needsPodExec.value }}
 - apiGroups: [""]
   resources: [pods/exec]
   verbs: [create]
-{{- end }}
+{{ end }}
+{{- if $needsNodesList }}
+{{- $nodeVerbs := list "list" -}}
+{{- if eq .Values.engine.name "k8s" -}}
+{{- $nodeVerbs = list "get" "list" "update" -}}
+{{- end -}}
+{{- if eq .Values.engine.name "slinky" -}}
+{{- $nodeVerbs = append $nodeVerbs "patch" -}}
+{{- end -}}
 - apiGroups: [""]
   resources: [nodes]
-  verbs: [get,list]
-{{- if eq .Values.engine.name "k8s" }}
-- apiGroups: [""]
-  resources: [nodes]
-  verbs: [update]
-{{- end }}
-{{- if eq .Values.engine.name "slinky" }}
-- apiGroups: [""]
-  resources: [nodes]
-  verbs: [patch]
-{{- end }}
+  verbs: {{ $nodeVerbs | toJson }}
+{{ end }}
+{{- if $needsDaemonsets }}
 - apiGroups: [apps]
   resources: [daemonsets]
   verbs: [get]
-{{- if eq .Values.engine.name "slinky" }}
+{{ end }}
+{{- if $needsConfigMaps }}
 - apiGroups: [""]
   resources: [configmaps]
   verbs: [create,get,update]
-{{- end }}
+{{ end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -73,6 +82,7 @@ roleRef:
   kind: ClusterRole
   name: {{ include "topograph.rbacName" . }}
   apiGroup: ""
+{{- end }}
 {{- if eq .Values.engine.name "nfd" }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/charts/topograph/tests/__snapshot__/render_snapshot_test.yaml.snap
+++ b/charts/topograph/tests/__snapshot__/render_snapshot_test.yaml.snap
@@ -420,11 +420,6 @@ renders default values.yaml:
         verbs:
           - get
           - list
-      - apiGroups:
-          - ""
-        resources:
-          - nodes
-        verbs:
           - patch
   15: |
     apiVersion: rbac.authorization.k8s.io/v1
@@ -1006,11 +1001,6 @@ renders values.k8s.gateway-api-example.yaml:
         verbs:
           - get
           - list
-      - apiGroups:
-          - ""
-        resources:
-          - nodes
-        verbs:
           - patch
   16: |
     apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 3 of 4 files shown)


dmitsh:
@Shankar-v27 , sorry for late response.
Could you please rebase your branch and resolve conflicts?

Shankar-v27:
Thanks Dmitsh! I’ve addressed the latest review feedback as well and pushed the additional test assertion fixes in a70d185. The changes are limited to strengthening the RBAC negative assertions, with the commit DCO- and SSH-signed.

Shankar-v27:
Thanks @dmitsh👀 I’ve seen the latest CodeRabbit comment as well. The current checks are passing, so I’ll leave the fixture unchanged for now unless you’d prefer me to make the topology configuration explicit.

dmitsh:
@Shankar-v27 , the chart-test failed, could you take a look?

charts/topograph/templates/rbac.yaml
```diff
@@ -25,40 +25,49 @@ already supplies nodes, a pod selector, or the default flat topology.
...
+{{ end }}
+{{ if $needsPodExec.value }}
 - apiGroups: [""]
   resources: [pods/exec]
   verbs: [create]
+{{ end }}
+{{ if $needsNodesList }}
+{{- $nodeVerbs := list "list" -}}
+{{- if eq .Values.engine.name "k8s" -}}
+{{- $nodeVerbs = list "get" "list" "update" -}}
```
dmitsh: **[P1] Grant `patch`, not `update`, to the Kubernetes engine.** The engine publishes labels with `CoreV1().Nodes().Patch`, and Kubernetes RBAC `update` does not authorize `patch`; label publication will therefore fail with HTTP 403. Please render `get`, `list`, and `patch` here and update the corresponding tests and snapshots.

charts/topograph/templates/rbac.yaml
```diff
@@ -25,40 +25,49 @@ already supplies nodes, a pod selector, or the default flat topology.
...
+{{- end -}}
+{{- if eq .Values.engine.name "slinky" -}}
```
dmitsh: **[P2] Gate Slinky `patch` on `useDynamicNodes`.** The Node patch call is reachable only from dynamic-node reconciliation, while non-dynamic Slinky needs only Node reads. Appending `patch` for every Slinky configuration retains an unnecessary cluster-wide write permission and conflicts with this PR’s least-privilege goal.

charts/topograph/tests/__snapshot__/render_snapshot_test.yaml.snap
```diff
@@ -47,7 +47,7 @@ renders default values.yaml:
       template:
         metadata:
           annotations:
-            checksum/config: 783f0643aeb7b18608d2eeb5fae26a211f3bddcc6e5c564da5c1f9f5fb03cf74
+            checksum/config: b883071c84f7fede8c950dc81c66e785e4e0516bf51254a1402be5ac099bd90c
```
ArangoGutierrez: The `checksum/config` values in this snapshot no longer match what the chart renders, so `helm unittest charts/topograph` fails with 24 snapshot mismatches (main is 153/153 green). Those hashes are the sha256 of the rendered configmaps, which `rbac.yaml` cannot influence, so they look regenerated against a tree carrying an unrelated change. Reverting the checksum lines to their base values or regenerating from a clean checkout should clear it.

charts/topograph/tests/rbac_test.yaml
```diff
@@ -183,6 +189,8 @@ tests:
 
   - it: does not grant node patch access to the slurm engine
     set:
+      provider:
```
ArangoGutierrez, line 267: Collapsing the node rules into one rule with computed verbs makes this test's only assertion unreachable: helm-unittest `notContains` matches the whole rule map, and the template can now only emit `[list]`, `[list, patch]` or `[get, list, patch]`, never a standalone `[patch]`. Mutating `$nodeVerbs` to include `patch` leaves this test green while five others go red. Asserting the exact rendered verb list (or `any: true`) restores the guard.

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Security
 
 - Helm now requires an explicit ServiceAccount name when creation is disabled for Topograph, node-observer, or node-data-broker, preventing cluster-scoped RBAC from being silently bound to the namespace's default ServiceAccount.
+- Gated Topograph API server RBAC ClusterRole rules dynamically to render permissions only when required by the configured provider or engine, collapsed node rules into a single rule with dynamically computed verbs, and omitted ClusterRole and ClusterRoleBinding entirely when no Kubernetes permissions are required ([#386](https://github.com/NVIDIA/topograph/issues/386)).
```
ArangoGutierrez: This entry landed in the `[v1.0.0]` section, which was released 2026-08-18, rather than under `[Unreleased]`. Probably a casualty of the merge from main that introduced the version heading.

ArangoGutierrez (commented):
Four notes, one of which is currently red in CI.

1. [must-fix] The `checksum/config` values in this snapshot no longer match what the chart renders, so `helm unittest charts/topograph` fails with 24 snapshot mismatches (main is 153/153 green). Those hashes are the sha256 of the rendered configmaps, which `rbac.yaml` cannot influence, so they look regenerated against a tree carrying an unrelated change. Reverting the checksum lines to their base values or regenerating from a clean checkout should clear it. (charts/topograph/tests/__snapshot__/render_snapshot_test.yaml.snap:50)
2. [should-fix] Collapsing the node rules into one rule with computed verbs makes this test's only assertion unreachable: helm-unittest `notContains` matches the whole rule map, and the template can now only emit `[list]`, `[list, patch]` or `[get, list, patch]`, never a standalone `[patch]`. Mutating `$nodeVerbs` to include `patch` leaves this test green while five others go red. Asserting the exact rendered verb list (or `any: true`) restores the guard. (charts/topograph/tests/rbac_test.yaml:192)
3. [should-fix] This entry landed in the `[v1.0.0]` section, which was released 2026-08-18, rather than under `[Unreleased]`. Probably a casualty of the merge from main that introduced the version heading. (CHANGELOG.md:70)
4. [consider] Switching these guards from `{{- if }}` to `{{ if }}` leaves blank lines between every rule in the rendered ClusterRole (and a few trailing ones after the last rule). Still valid YAML, just noisier than the rest of the chart. (charts/topograph/templates/rbac.yaml:39)

dmitsh:
Thank you @Shankar-v27

dmitsh (approved):

charts/topograph/templates/rbac.yaml
```diff
@@ -65,14 +65,16 @@ rules:
   verbs: [create]
 {{- end }}
 {{- if $needsNodes }}
+{{- $nodeVerbs := list "list" -}}
+{{- if eq $engine "k8s" -}}
+{{- $nodeVerbs = list "get" "list" "patch" -}}
+{{- end -}}
+{{- if and (eq $engine "slinky") (get (default dict .Values.engine.params) "useDynamicNodes") -}}
```
ArangoGutierrez, line 72: The gate reads engine.params.useDynamicNodes at install time, but that is a documented per-request field (docs/api.md:104) and readRequest only backfills config params the payload omits, so a request setting it to true against a release installed without it reaches Nodes().Patch and the 403 surfaces to the caller as a 502. It also fails the other way: helm template --set-string engine.params.useDynamicNodes=false renders ["list","patch"], because the template tests raw truthiness while the engine decodes the value through ParseBool.

ArangoGutierrez (commented):
Nothing here blocks merge.

The verb matrix matches what the code actually does: the k8s engine is the only path that calls Nodes().Get and Nodes().Patch, and slinky only patches inside performReconciliation, which engine.go:488 gates on UseDynamicNodes, so dropping get and patch elsewhere is right. nfd, dra and infiniband-k8s all reach nodes through internal/k8s.GetNodes, which is a List, so [list] fits them. make chart-test is green at this head, and widening the default node verbs to [list, update] does fail the suite. One thing on the slinky gate is inline. Worth fixing the PR title as well: it still carries the branch name rather than describing the RBAC change, and that text lands in the squash commit.

1. The notContains guards on the nodes rule cannot fail: helm-unittest compares whole rule entries, and the template only ever emits [list], [list, patch] or [get, list, patch]. Widening the default verbs to [list, update] failed the suite entirely through the contains asserts, with no notContains firing, including the one aimed at update. (charts/topograph/tests/rbac_test.yaml:79)

Shankar-v27:
> Thank you @Shankar-v27

@dmitsh @ArangoGutierrez Thank you for your continuous guidance, means a lot !!

Shankar-v27:
@dmitsh I have a small favour to ask that's easier over email than here. Could you share your email, or would you be open to me reaching out via LinkedIn?

Thanks!

PR #446, opened 2026-08-03
refactor(node-observer): reconcile topology with workqueue

### LINKED ISSUE #445: Refactor node-observer into a workqueue-based reconciliation controller
## Problem

The node-observer currently coordinates topology generation through an ad hoc
buffered channel and mutable retry timer. Although this coalesces some events,
it does not follow the standard Kubernetes controller reconciliation pattern
and makes retry, shutdown, and event-handling behavior difficult to reason
about.

The refactor must preserve the topology-request trigger behavior implemented
on the current `main` branch. Expanding the set of triggering events, including
adding Node update handling or an unconditional startup request, is outside the
scope of this change.

## Proposed change

Refactor the node-observer to use a client-go rate-limiting workqueue and an
idempotent reconciliation cycle.

Because topology generation is cluster-wide, all existing relevant events
should enqueue a single topology key rather than one key per Kubernetes object.

The controller should:

- Preserve the Node, Pod, API-server, and broker event predicates from `main`.
- Wait for configured informer caches to synchronize before processing work.
- Avoid an additional synthetic reconciliation after cache synchronization;
  initial informer Add events remain unchanged.
- Coalesce bursts of events into a single topology reconciliation.
- Delay reconciliation while the node-data-broker DaemonSet is not ready.
- Rate-limit retries after topology-generation request failures using the
  configured retry delay.
- Allow a newer informer event to reconcile immediately and supersede any
  pending retry from an older attempt.
- Prevent obsolete delayed workqueue entries from issuing redundant topology
  requests.
- Cancel in-flight requests and shut down the queue cleanly on termination.

## Acceptance criteria

- The channel/timer retry loop is replaced by a client-go rate-limiting
  workqueue.
- Reconciliation uses one cluster-wide topology key.
- Topology-request trigger conditions match the current `main` branch:
  - Node add and delete events enqueue reconciliation.
  - Node update events do not enqueue reconciliation.
  - Existing Pod, API-server, and broker predicates remain unchanged.
- Controller startup does not issue a topology request without an informer
  event.
- Duplicate events are coalesced.
- Failed requests are requeued using the configured retry delay.
- A newer event supersedes and resets a pending retry from an older attempt.
- Obsolete delayed entries do not issue requests after a newer successful
  reconciliation.
- Broker readiness causes a delayed requeue without counting as a topology
  generation failure.
- Informer cache synchronization failures and controller shutdown are handled
  explicitly.
- Unit tests cover event-trigger parity, deduplication, retries, superseded
  delayed entries, broker readiness, cache cancellation, and shutdown.
- Architecture documentation and the changelog are updated.

## Non-goals

- Changing or expanding the events that trigger topology generation.
- Adding Node update predicates.
- Performing unconditional or periodic topology reconciliation.
- Converting the node-data-broker into a periodic node-local reconciliation
  agent.
- Removing the existing cluster-wide broker readiness gate.
- Introducing a CRD, leader election, or controller-runtime dependency.
- Changing the Provider → canonical Graph → Engine boundary.

Changes to trigger behavior or periodic reconciliation can be considered
independently after the node-observer has a standard reconciliation foundation.

dmitsh:
Replace the ad hoc channel and timer loop with a client-go
rate-limiting workqueue using one cluster-wide topology key.

Coalesce event bursts, wait for informer caches to synchronize,
retry failed generation requests, and requeue while the node-data
broker is not ready.

Add coverage for queue deduplication, retries, cache cancellation,
broker readiness, and Node update predicates.

closes #445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- The node-observer now processes its existing topology-generation triggers through a client-go rate-limiting work queue, coalescing event bursts into a single cluster-wide reconciliation while preserving existing trigger and retry behavior.
 - **BREAKING:** Simulation models now define accelerator topology through inherited `switches[].annotations` and `blocks[].annotations` using `accelerator.topology.test/domain` and optional `accelerator.topology.test/sub-domain`, instead of the `network.topology.nvidia.com/accelerator` model label.
 - The Slurm topology-update trigger script now accepts AWS, GCP, OCI, Nebius, NetQ, Nscale, Lambda, and bare-metal InfiniBand providers.
 - **BREAKING:** Fabric topology now uses variable, closest-first tiers labeled `network.topology.nvidia.com/tier-N`; `InstanceTopology.FabricTiers` and graph conversion support arbitrary fabric depth. Accelerator topology remains a single `AcceleratorID`/`Graph.Domains` dimension labeled `network.topology.nvidia.com/accelerator`. The fixed `leaf`, `spine`, and `core` keys and process-wide Helm/CLI label overrides are replaced by optional `fabricLabels` and `acceleratorLabel` parameters on the `k8s` engine.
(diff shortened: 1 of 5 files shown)


faganihajizada (approved):
Looks good!

Out of scope but noticed that obsolete delayed entries are discarded via `completedGeneration`. check `processNextWorkItem` and `TestSuccessfulEventSupersedesPendingRetry`.

PR #449, opened 2026-08-05
feat: nscale api change

ravisoundar:
## Description
Updates the nscale provider to use the `list-placement-servers` end point instead of the `instances` api.

## Checklist

### DIFF
--- a/pkg/providers/nscale/imds.go
+++ b/pkg/providers/nscale/imds.go
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2026 NVIDIA CORPORATION
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package nscale
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"k8s.io/klog/v2"
+
+	"github.com/NVIDIA/topograph/internal/exec"
+	"github.com/NVIDIA/topograph/pkg/providers"
+	"github.com/NVIDIA/topograph/pkg/topology"
+)
+
+const (
+	IMDSURL = "http://169.254.169.254/openstack/latest/meta_data.json"
+
+	metaKeyServerID = "serverID"
+	metaKeyRegionID = "regionID"
+)
+
+type imdsMetadata struct {
+	Meta map[string]string `json:"meta"`
+}
+
+func pdshCmd(url string) string {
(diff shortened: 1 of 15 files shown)


pkg/providers/nscale/imds.go
```diff
@@ -0,0 +1,164 @@
...
+	metaKeyServerID = "serverID"
+	metaKeyRegionID = "regionID"
+)
+
+type imdsMetadata struct {
+	Meta map[string]string `json:"meta"`
+}
+
+func pdshCmd(url string) string {
+	return fmt.Sprintf("echo $(curl -fsS -- %s)", shellQuote(url))
```
dmitsh: **[P1] Preserve the `curl` failure status**

Wrapping `curl` in `echo $(...)` masks the remote failure: the shell returns `echo`'s exit status (0) even when `curl -f` cannot reach IMDS. In an all-node IMDS outage, `PdshTolerant` therefore sees success, parsing produces an empty mapping, and Slurm generation can replace `topology.conf` with no topology. Please structure the command so `curl` failure remains nonzero (for example, assignment followed by `&&` before emitting the response).

pkg/providers/nscale/provider.go
```diff
@@ -22,44 +24,62 @@ import (
 const (
 	NAME = "nscale"
 
-	urlTopologyPath  = "/v1/topology"
-	urlInstancesPath = "/v2/instances"
+	urlTopologyPath = "/v2/topology"
```
dmitsh, line 27: **[P2] Align the documented Radar endpoint**

This changes the implementation to `/v2/topology`, but `docs/providers/nscale.md` still tells operators to use `/v1/topology` in both the data-source summary and the request example. Please update those references to `/v2/topology` (or revert this path if v1 remains the intended contract).

pkg/providers/nscale/provider.go
```diff
@@ -69,15 +89,6 @@ type TopologyResult struct {
...
-type instance struct {
-	Metadata instanceMetadata `json:"metadata"`
-}
-
-type instanceMetadata struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
-}
-
 func (c *nscaleClient) Topology(ctx context.Context, region string, pageSize, offset int) ([]InstanceTopology, error) {
```
dmitsh: Topology() always returns httpErr. Let's update the signature.
That would make type checking redundant in baseProvider.generateRegionInstanceTopology()

dmitsh (approved):

PR #456, opened 2026-08-10
feat(k8s): use vendor-neutral topology label domains

dmitsh:
Replace the NVIDIA-specific fabric and accelerator label namespaces with
fabric.topograph.run and accelerator.topograph.run.


ravisoundar (approved):
lgtm

PR #457, opened 2026-08-11
fix(infiniband): make accelerator discovery pluggable

### LINKED ISSUE #458: [FEA]: Make accelerator-domain discovery pluggable
Providers currently couple network-fabric discovery with accelerator-domain
discovery. For example, the InfiniBand providers use `ibnetdiscover` for the
network fabric while also containing NVIDIA-specific logic for discovering
NVL domains.

This coupling makes it difficult to:
- Reuse the same accelerator discovery with another network-fabric provider.
- Use a different accelerator-domain source with the InfiniBand provider.
- Disable accelerator discovery without affecting fabric discovery.
- Add new accelerator sources without adding provider-specific switches.

The existing `useGpuCliqueLabel` boolean also does not scale beyond choosing
between two implementations and assumes a specific Kubernetes label.

**Describe the solution you'd like**

Introduce a pluggable accelerator-domain discovery interface that providers
can compose with their existing fabric discovery.

Add an optional provider parameter:

```yaml
accelerator:
  source: kubernetes-label
  kubernetesLabel:
    key: example.com/accelerator-domain
```

Initially support:
- nvidia-smi
- kubernetes-label
- none

Omitting accelerator, specifying accelerator: {}, or selecting
source: none should disable accelerator discovery. Any non-empty section
should require an explicit, valid source.

For `nvidia-smi`, query `fabric.clusterUuid` and `fabric.cliqueId`, validate
the results, and expose the accelerator-domain ID as `ClusterUUID.CliqueId`.

The Kubernetes deployment should grant GPU Operator lookup and pod-exec
permissions only when the `nvidia-smi` source is selected.

dmitsh:
Decouple accelerator-domain discovery from InfiniBand fabric discovery.
Support nvidia-smi, Kubernetes Node labels, and disabled discovery through
provider.params.accelerator.

Query NVL partition IDs using the nvidia-smi CSV interface, deduplicate
per-GPU results, reject unavailable fields, and normalize IDs as
ClusterUUID.CliqueId. Grant node-data-broker GPU Operator permissions only
when nvidia-smi discovery is enabled.

BREAKING CHANGE: remove useGpuCliqueLabel. Omitting accelerator, using an
empty accelerator section, or selecting source: none disables accelerator
discovery.

closes #458

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -18,6 +18,8 @@ Topograph discovers the physical network topology of a cluster (NVLink domains, 
 
 Providers differ by environment. The canonical `topology.Graph` is stable. Engines only translate — they do not discover.
 
+Within a provider, network-fabric and accelerator-domain discovery may be composed independently through `pkg/accelerator`; the provider remains responsible for combining both dimensions into the canonical graph.
+
 This separation is load-bearing. If you find yourself reading the fabric in an engine, or emitting scheduler-specific output from a provider, stop and reconsider.
 
 ### Repository map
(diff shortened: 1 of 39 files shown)


ravisoundar (dismissed):
lgtm

PR #471, opened 2026-08-17
fix(provider/lambdai): name the real credential problem

alaski-lambda:
Loading the provider reported the wrong missing key whenever workload identity had not been injected. requireStaticCredentials validated workspaceId from the credentials map only, and ran before the credentials-then-params resolution below it, so a workspaceId supplied as a provider parameter -- the arrangement the workload-identity docs recommend -- was reported as `missing 'workspaceId'` while sitting plainly in the operator's values file. The real problem in that case is that no token was supplied and no identity was injected.

Resolve the workspace from credentials-then-params before the mode branch and validate it once, so the same message covers both modes and names both places. requireStaticToken now checks only the token, and distinguishes a named-but-blank token -- a malformed credential, with a hint to omit the key entirely when the pod does carry an identity -- from no token at all.

That second state is ambiguous: a Slurm or bare-metal deployment that forgot its token looks exactly like a Kubernetes pod whose workload identity was never injected. The error therefore leads with the token every deployment can supply, and offers the LAMBDA_ROLE_LRN env var, the lambda.ai/role-lrn ServiceAccount annotation, and the webhook as the workload-identity alternative, rather than presenting Kubernetes troubleshooting to a Slurm operator as the only path.

Both workspace sources are trimmed before selection, preserving the TrimSpace the static credential check applied and extending it to the parameter, so a whitespace-only value counts as absent rather than being forwarded verbatim as the workspace_id query value. A blank workspace still falls through to the other source rather than being rejected outright the way a blank token is: the workspace selects no principal, so falling through cannot change who a request authenticates as.

This also collapses the `parameters error:` / `credentials error:` split that reported the same missing key two different ways depending on the authentication mode, and makes static-token mode accept a workspaceId parameter as workload-identity mode already did -- one resolution rule for both modes. No release has shipped either behavior, so there is no CHANGELOG entry.

## Checklist

### DIFF
--- a/pkg/providers/lambdai/provider.go
+++ b/pkg/providers/lambdai/provider.go
@@ -208,12 +208,36 @@ func Loader(ctx context.Context, config providers.Config) (providers.Provider, *
 	identityLRN := os.Getenv(envRoleLRN)
 	wiMode := !tokenSupplied && identityLRN != ""
 
+	// The workspace is required by both modes and may be supplied as a credential
+	// or, since it is an identifier rather than a secret, as a provider parameter.
+	// Resolve it before the mode branch so a missing value reports the same error
+	// either way: validating it inside the static credential check reported
+	// workspaceId as missing whenever workload-identity injection had not
+	// happened, even when it was set in params, sending operators after the wrong
+	// key.
+	//
+	// Both sources are trimmed before selection, so a blank value counts as
+	// absent and the surviving one reaches the API without surrounding
+	// whitespace, which would otherwise be sent verbatim as workspace_id.
+	// Unlike a blank token, a blank workspace is not treated as a malformed
+	// credential to be rejected outright: the workspace selects no principal, so
+	// falling through to the other source cannot change who a request
+	// authenticates as.
+	workspaceID := strings.TrimSpace(creds.WorkspaceID)
+	if workspaceID == "" {
+		workspaceID = strings.TrimSpace(params.WorkspaceID)
+	}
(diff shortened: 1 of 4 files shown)


pkg/providers/lambdai/provider.go
```diff
@@ -208,12 +215,36 @@ func Loader(ctx context.Context, config providers.Config) (providers.Provider, *
...
+	// Both sources are trimmed before selection, so a blank value counts as
+	// absent and the surviving one reaches the API without surrounding
+	// whitespace, which would otherwise be sent verbatim as workspace_id.
+	// Unlike a blank token, a blank workspace is not treated as a malformed
+	// credential to be rejected outright: the workspace selects no principal, so
+	// falling through to the other source cannot change who a request
+	// authenticates as.
+	workspaceID := strings.TrimSpace(creds.WorkspaceID)
+	if workspaceID == "" {
+		workspaceID = strings.TrimSpace(params.WorkspaceID)
```
dmitsh: is there a reason to allow workspaceID in both "params" and "creds"?

alaski-lambda: No strong reason. It was done in an attempt to keep the workload-identity option "secret free". However trying to support these two paths is more confusing than beneficial. I'll open another PR to clean this up.

alaski-lambda: Changed approach and added it to this PR.

dmitsh (approved):

PR #486, opened 2026-08-27
test: add coverage for httperr and registry packages

resker:
## Description

Adds test files for two packages that had zero coverage.

### `internal/httperr`

`NewError`, `Error`, and `Code` were uncovered despite being called at every provider/engine boundary. The new test exercises all three methods across several HTTP status codes, bringing the package to 100%.

### `pkg/registry`

The central provider and engine wiring had no tests. The new test asserts that every registered name resolves to a non-nil loader, and that unregistered names return a `400 Bad Request` error — covering the registry initialization, the `Get` happy path for every provider and engine, and the `Get` error path in both `providers.Registry` and `engines.Registry`.

## Checklist

- [x] `make qualify` passes (runs fmt, vet, lint, test)
- [x] New or changed public behavior is covered by a test
- [x] Documentation impact evaluated — no doc changes needed
- [x] User-facing changes recorded in `CHANGELOG.md` `[Unreleased]` when applicable — tests only
- [ ] `pkg/topology/` changes were discussed in an issue first — N/A
- [x] Every commit has a DCO sign-off

### DIFF
--- a/internal/httperr/http_error_test.go
+++ b/internal/httperr/http_error_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


pkg/registry/registry_test.go
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
dmitsh: 2026

internal/httperr/http_error_test.go
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
dmitsh: 2026

dmitsh (approved):

PR #487, opened 2026-08-27
feat(provider/crusoe): add the Crusoe Cloud topology provider

afuller-crusoe:
## Description

Adds a topology provider for [Crusoe Cloud](https://crusoe.ai/), plus its simulation variant.

Crusoe compute is virtualized. A guest VM cannot run `ibnetdiscover` against the InfiniBand fabric, and there is no per-node metadata service that reports switch identity — so neither `infiniband-k8s` nor an IMDS-style provider works. The Crusoe control plane instead publishes each VM's fabric position as Kubernetes Node labels, and this provider reads them.

**Fabric tiers**, closest-first:

| Tier | Source label | Meaning |
|---|---|---|
| 0 | `crusoe.ai/pod.id` | Rail-optimized pod — hosts wired to a shared set of InfiniBand leaf switches, one per GPU rail |
| 1 | `crusoe.ai/ib.partition.id` | InfiniBand partition — the isolated fabric the pod belongs to |
| 2 | `crusoe` | A synthetic root above every partition |

The root tier is not a switch. Slurm needs one common ancestor so a job can span partitions, and so that nodes with no InfiniBand still belong to the tree. Nodes carrying neither label get placeholder tiers (`cpu-pod` / `cpu-partition`) beneath the same root and stay schedulable in one tree.

The provider supplies the switch fabric hierarchy, for use with `topology/tree`.

**Instance identity**: the provider takes the instance ID from the `topograph.run/instance` annotation and falls back to the node name. Engines key their `ComputeInstances` map on that annotation, so a provider that assumes the two are equal silently resolves nothing in the graph wherever they differ.

### Testing

Unit tests cover label extraction, tier construction, the fallback for nodes without InfiniBand labels, request filtering, multi-region rejection, and the rendered `topology.conf` for both the live and simulated providers. Package coverage is 83.5%.

`.github/workflows/k8s-test.yaml` gains a `crusoe` case. It renders `tests/models/crusoe-small.yaml` into KWOK Nodes — which carry the Crusoe labels, since model switch labels propagate to their nodes — and runs the **live** provider against them, so CI exercises the production code path rather than only the simulator. The runner's expected fabric tiers and accelerator domain are now parameterized instead of hardcoded to `medium.yaml`, with the previous values as defaults so the existing matrix entries are unchanged.

Verified on kind + KWOK. Resulting labels:

```
NAME        tier-0                                tier-1                                tier-2
gpu-01/02   9b1c7d54-3e08-4a92-8f61-2d5a4c7b0e13  0cf47922-9f9e-4a2e-b187-5c51697e7739  crusoe
gpu-03/04   4f6a2e91-8d75-4c30-b1e8-6a9c3f5d7b24  0cf47922-9f9e-4a2e-b187-5c51697e7739  crusoe
gpu-05/06   7c3d9f68-5b21-4e84-a0f3-9e1b8d2c6a57  1d0e8b3a-2c47-4f16-9a55-7be3d9c4f082  crusoe
cpu-01/02   cpu-pod                               cpu-partition                         crusoe
```

Two rail-optimized pods under one partition, a second partition, and the no-InfiniBand fallback — all under one root.

Also validated against a live Crusoe cluster with InfiniBand-attached GPU nodes and a separate CPU nodepool. `scontrol show topology` reports the full three-level hierarchy, the root spans both the InfiniBand and non-InfiniBand nodes, and `srun -N2 --switches=1` places the job on the only pair of nodes sharing a tier-0 grouping.

### Notes for reviewers

- `nodeSelector` has no default. Without one, every Node in the cluster is included and nodes with no InfiniBand labels land under the placeholder tiers. This matches the other Kubernetes providers; the provider doc calls it out.
- The chart change is narrow: `crusoe` added to the `provider.name` enum, to the Node-list RBAC gate, and to the values comment.
- `docs/providers/crusoe.md`, `docs/index.yml`, `docs/overview.md` (provider list and the Choosing a Provider table), and `CHANGELOG.md` are updated. `AGENTS.md` and `.claude/CLAUDE.md` get `crusoe` in the repository-map provider list.

## Checklist


dmitsh:
Thank you for contributing, @afuller-crusoe ,
Does this topology provider also generate accelerator domain (NVL partition) labels?

pkg/providers/crusoe/provider.go
```diff
@@ -0,0 +1,133 @@
...
+
+// nodeMetadataFromNodeList reads the instance identity the node-data-broker
+// wrote and falls back to the Kubernetes node name. Engines key their
+// ComputeInstances map on that annotation, so the provider has to agree with it
+// or nothing resolves in the graph. On Crusoe the two are the same string, but
+// deferring to the annotation keeps the provider correct if that ever changes.
+func nodeMetadataFromNodeList(nodeList *corev1.NodeList) []nodeMetadata {
+	nodes := make([]nodeMetadata, 0, len(nodeList.Items))
+	for _, node := range nodeList.Items {
+		instanceID := strings.TrimSpace(node.Annotations[topology.KeyNodeInstance])
```
ArangoGutierrez, line 101: Replacing this whole annotation read with a plain `instanceID = node.Name` keeps the package suite green, so the identity rule the description calls out has no test behind it. Every Node built by the `node` helper in provider_test.go has nil annotations, so only the fallback branch ever runs. A Node whose `topograph.run/instance` differs from its name, requested under that annotation, would pin it.

pkg/registry/registry.go
```diff
@@ -40,6 +41,8 @@ import (
 var Providers = providers.NewRegistry(
 	aws.NamedLoader,
 	aws.NamedLoaderSim,
+	crusoe.NamedLoader,
```
ArangoGutierrez, line 44: `crusoe` and `crusoe-sim` are missing from the `known` list in registry_test.go, so deleting both loader entries here leaves `pkg/registry` passing. Every other provider is listed there, sim variants included, which is what makes that test the guard on this file.

docs/providers/crusoe.md
```diff
@@ -0,0 +1,119 @@
...
+| 0 | `crusoe.ai/pod.id` | Rail-optimized pod — hosts wired to a shared set of InfiniBand leaf switches, one per GPU rail |
+| 1 | `crusoe.ai/ib.partition.id` | InfiniBand partition — the isolated fabric the pod belongs to |
+| 2 | `crusoe` | A synthetic root above every partition |
+
+The root tier is not a switch. Slurm needs one common ancestor so a job can
+span partitions, and so nodes with no InfiniBand still belong to the tree.
+
+### Nodes without InfiniBand
+
+Nodes that carry neither label get the placeholder tiers `cpu-pod` and
```
ArangoGutierrez, line 43: `readFabricLabels` returns false when either label is missing rather than only when both are, so a GPU node carrying `crusoe.ai/ib.partition.id` with no `crusoe.ai/pod.id` lands under `cpu-pod` and `cpu-partition` beside the CPU nodepool. The behavior looks deliberate given the partition without a pod is not enough case, so it is the wording that wants to be `either` instead of `neither`.

Also at pkg/providers/crusoe/k8s.go:44.

ArangoGutierrez (commented):
Thanks for this. Hanging the CPU placeholder tiers off the same synthetic root is what keeps a mixed InfiniBand and CPU cluster inside one Slurm tree, and the golden topology.conf test pins that shape precisely.

The provider stays on the right side of the discovery and translation split, and the rendered topology.conf assertions pin the tier shape rather than graph internals.

- The clientset is built straight from `rest.InClusterConfig()` with no `k8s.ConfigureClientRateLimits(cfg)` in between, so an operator setting `kubeClient.qps` gets `KUBE_QPS` onto the pod where it then does nothing for this provider's Node lists. The dra provider and the k8s, nfd and slinky engines all call it; `infiniband-k8s` is the one that also skips it. (pkg/providers/crusoe/provider.go:61)
- This is the one doc surface that did not move with the rest. Its per-provider table is exhaustive over Kubernetes-engine providers, with dra the single omission and its reason written out below the table, so `crusoe` wants a No / Yes row next to `nebius`. The two `crusoe.ai` keys also fit the Without Topograph table alongside the `oci.oraclecloud.com` and `cloud.google.com` entries. (docs/reference/node-labels.md:40) https://github.com/NVIDIA/topograph/blob/d22ae99b46a3c62122820f50495fa61ea6581efa/docs/reference/node-labels.md#L34-L44

ArangoGutierrez (approved):
The provider reads the Crusoe control plane labels and stops there, leaving translation to the engines, and the simulation variant runs the same buildClusterTopology path as the live one rather than a parallel implementation of it.

dmitsh (approved):

PR #496, opened 2026-08-31
feat(provider/crusoe): group blocks by NVLink domain

afuller-crusoe:
## Description

Follow-up to #487. The Crusoe provider now derives accelerator domains from `nvidia.com/gpu.clique`, so `topology/block` renders one Slurm block per NVLink domain.

Block scheduling only pays off if a block maps to the fabric a job wants to stay inside. On rack-scale NVLink systems such as GB200 and GB300 NVL72 that fabric is the NVLink domain, not the InfiniBand partition. GPU Feature Discovery advertises the domain as `nvidia.com/gpu.clique` with the value `<clusterUUID>.<cliqueID>`. Every node in a rack carries the same value, and a clique never spans racks.

**The InfiniBand partition is deliberately not used as a fallback domain.** A partition can span many racks while a clique cannot, so keying blocks on the partition would let Slurm spread one job across racks and quietly fall back to InfiniBand instead of NVLink — the opposite of what block topology is for. Nodes with no clique label get no accelerator domain and are scheduled by the switch tree alone, which leaves every existing GPU class unchanged.

A GPU SKU that ships without InfiniBand carries a clique but no fabric labels. It contributes a block anyway, so it is not silently dropped from block topology.

Rendered output:

```
# block001=nvl-<clusterUUID>.32766
BlockName=block001 Nodes=gpu-[01-02]
# block002=nvl-<clusterUUID>.32767
BlockName=block002 Nodes=gpu-[03-04]
BlockSizes=2,4
```

The `nvl-` prefix marks the domain as NVLink-derived so the source of a block is readable in the accelerator topology.

### Testing

Unit tests cover clique extraction, the no-clique case, a node with a clique but no fabric labels, and the rendered `topology.conf` for both tree and block plugins. Package coverage is 84.3%.

The block test pins the rendered file rather than only the intermediate graph, so a regression in domain selection shows up as a file diff. It also pins emission order, which matters beyond readability: unstable ordering would rewrite the topology ConfigMap on every regeneration and churn `slurmctld` reconfigures.

### Notes for reviewers

- `docs/reference/node-labels.md` had no `crusoe` row at all — an omission in #487. This adds one recording both capabilities: accelerator domains optional (present only on NVLink hosts), fabric tiers always.
- No change to the fabric tiers or to any node that lacks a clique label, so this is additive for existing deployments.

## Checklist


dmitsh:
@afuller-crusoe , thank you for the PR.
Please take a look at this code change, as it might affect your assumptions.
https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/pull/1247

afuller-crusoe:
@dmitsh Thanks for the feedback. 

I looked into the DRA driver changes. It seems this is more operational work on how we manage DRA driver (setting `gpuCliqueLabelEnabled=true` for backwards compatibility), rather than anything impacting the topograph provider. I looked into the DRA provider and others that support block topology and they all seem to rely on reading this label. Crusoe docs do need an update to specify the label comes from GFD or the DRA driver after ComputeDomain creation.

Is all this correct or am I missing something?

Thanks for raising this, it needs an internal ticket to track the backwards compatibility changes needed  to support this when upgrading to DRA driver 0.5.0, so glad it came up

dmitsh (approved):

dmitsh:
Hi @afuller-crusoe ,
We’ll be conducting an open source workshop at NVIDIA’s HQ on September 30. The goal is to deepen collaboration among contributors. Would you be interested in participating?

PR #510, opened 2026-09-04
kwok-nodes fix -model flag

kevin-kho:
## Description

closes #509 

## Checklist
- [x] I am familiar with the [Contributing Guidelines](./CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] All commits are signed off per DCO (`git commit -s`).

PR #511, opened 2026-09-04
fix(kwok-nodes): accept model basenames without the .yaml suffix

### LINKED ISSUE #509: [BUG]: kwok-nodes -model flag requires .yaml suffix
### Prerequisites

- [x] I searched existing issues and found no duplicates
- [x] I can reproduce this issue consistently

### Bug Description

In the DEVELOPMENT.md example, it shows how you can use kwok-nodes to generate example nodes.

The example given is `./bin/kwok-nodes -model small-tree -output -`

This fails with the following: `failed to read small-tree: open models/small-tree: file does not exist`

This is due to kwok-observer trying to open the file at the path: `/tests/models/small-tree`, without the `.yaml` suffix

If you rerun with the `.yaml` suffix: `./bin/kwok-nodes -model small-tree.yaml -output -`, it will succeed.

There needs to be a check for any `'.yaml` or `'.yml` suffix before attempting to open the path. 

### Component

Other / Unknown

### Steps to Reproduce

1. Run `./bin/kwok-nodes -model small-tree -output -`
2. The command will error out.
3. Run  `./bin/kwok-nodes -model small-tree.yaml -output -`
4. The command will succeed

### Environment

- Topograph version: main
- Deployment method (binary/deb/rpm/container/Helm): binary
- Scheduler (Slurm/Kubernetes/Slinky): N/A
- Cluster environment (bare-metal/CSP, specify provider): N/A

### Configuration

```yaml

```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

### Code of Conduct

- [x] I agree to follow Topograph's Code of Conduct

iacker:
Fixes #509

`kwok-nodes -model small-tree` fails with `open models/small-tree: file does not exist`, while the flag help says basenames resolve from `tests/models`. `GetModelFileData` now appends `.yaml` when the basename has no extension, so both the documented form and `small-tree.yaml` work. Paths with a directory component are unchanged.

Covered by a small test in `pkg/models` and checked with `./bin/kwok-nodes -model small-tree -output -`.

### DIFF
--- a/pkg/models/model_test.go
+++ b/pkg/models/model_test.go
@@ -21,6 +21,14 @@ func acceleratorDomainAnnotations(domain string) map[string]string {
 	return map[string]string{annotationAcceleratorDomain: domain}
 }
 
+func TestNewModelFromFileBasename(t *testing.T) {
+	for _, name := range []string{"small-tree", "small-tree.yaml"} {
+		cfg, err := NewModelFromFile(name)
+		require.NoError(t, err, name)
+		require.NotEmpty(t, cfg.Nodes, name)
+	}
+}
+
 func TestNewModelFromFileMedium(t *testing.T) {
 	cfg, err := NewModelFromFile("../../tests/models/medium.yaml")
 	require.NoError(t, err)
--- a/tests/model.go
+++ b/tests/model.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"embed"
 	"fmt"
+	"path/filepath"
 )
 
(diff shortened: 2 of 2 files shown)


dmitsh (changes requested):
To be consistent with the option to provide external model file, I would suggest fixing documentation and specifying filename with extension, like "./bin/kwok-nodes -model small-tree.yaml -output -"

dmitsh:
closed in favor of https://github.com/NVIDIA/topograph/pull/510

